PR #70, opened 2021-10-27
script update and schema fix

shunjizhan:
## Change
- fixed schema and mapping handler error so indexer and query can run.
- added script so it is more convenient to run separate services
- updated README

## Test
tested with local acala node, indexer is able to index data, and saw correct data from the graphql explorer


ThunderDeliverer (approved):
This looks good to me 👌 

@shunjizhan, would you mind looking and trying to run #72 and if it's ok, I'd like to merge it into this branch, before we merge this PR to master. That branch contains both linux and macOS Docker config and instructions on how to run which one.

Additionally I'd like to do a grammatical fix of the README once we are done with all of the changes and troubleshooting of the subql and adapter.

shunjizhan:
Thanks Jan, actually I might also need to modify the script and README back to using globally installed `subql/node` before merging it, since I found using it in local `node_modules` will cause it to fail to report some err msg somehow

ThunderDeliverer:
I added the newly released Mandala node to docker in #74, so I think we should merge that into this branch as well before merging this one.

shunjizhan:
## Change
- integrated helper function from `eth-provider`
- fixed schema and mapping handler bugs so indexer and query can run
- added script so it is more convenient to run separate services
- updated README with detailed instruction on how to run 

## Test
first cleaned up everything, and tested this branch locally following steps in README. Both docker and separate services work well, indexer can index EVM Txs and we can query them in subql playground.

xlc (approved):

PR #72, opened 2021-10-29
Modify Docker configuration to run on macOS

ThunderDeliverer:
Modified evm-subql docker configuration to be able to be run on macOS.
Additionally .env file example has been added to enable eth-rpc-adapter
to run on macOS.


evm-subql/project.yaml
```diff
@@ -3,7 +3,7 @@ description: ''
 repository: ''
 schema: ./schema.graphql
 network:
-  endpoint: ws://localhost:9944
+  endpoint: ws://host.docker.internal:9944
```
xlc, line 6: should revert this. I think it is possible to override endpoint for subquery via cli args

shunjizhan:
Tried the docker image and I got two errors,  wonder if I am missing anything?

when I run with `docker-compose -f macos-docker-compose.yml up`, I kept getting 
```
<indexer> ERROR Not same network: genesisHash different - 0x0fcddee6c619c40139a35703e9c3ad5b7fa972b2e0c307950fa74a3adcce04e7 : 0x775757cd9659db3eacfc8c72ac5a092625fb746ed73dc5bb2cf6dd4e95811fdb
```
Am I suppose to clear cache for the database in some way? I thought each docker run is already a fresh clean start without cache, so we don't need to clean cache manually? 

Also I am getting 
```
836fc88c4d42_evm-subql_graphql-engine_1 | Failed to fetch indexer health,  request to http://subquery-node:3000/health failed, reason: connect ECONNREFUSED 172.18.0.3:3000
```
is this expected behavior that will need a fix later?

evm-subql/README.md
```diff
@@ -20,7 +20,7 @@ yarn build
 ### run all services with docker
 for **linux users**, simply do `docker-compose up`, that's all. 
 
-for **mac users**, docker will need some different config, please checkout [this branch](https://github.com/AcalaNetwork/eth-rpc-adaptor/tree/mac-docker-settings/evm-subql) for running docker with mac.
+for **mac users**, use a macOS specfic docker compose with `docker-compose -f macos-docker-compose-yml up`.
```
shunjizhan, line 23: small typo `macos-docker-compose-yml`

xlc:
Docker volumes are persisted. You can do `docker-compose down && docker volume prune` to clean everything.

shunjizhan:
Thanks Bryan, after pruning docker volume I can run the this docker image in my Mac, and also able to reproduce the same error when I run a separate DB and  subql/node service. I am still investigating this error.
```
<fetch> ERROR failed to index block at height 370 handleEvmEvent([{"phase":{"applyExtrinsic":2},"event":{"index":"0xb402","data":["0x82a258cb20e2adb4788153cd5eb5839615ece9a0","0x0000000000000000000000000000000001000000",[{"address":"0x0000000000000000000000000000000001000000","topics":["0x8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925","0x00000000000000000000000082a258cb20e2adb4788153cd5eb5839615ece9a0","0x000000000000000000000000546411ddd9722de71da1b836327b37d840f16059"],"data":"0x00000000000000000000000000000000000000000000000000005af3107a4000"}]]},"topics":[]}])
subquery-node_1   | SequelizeDatabaseError: current transaction is aborted, commands ignored until end of transaction block
```

By the way I am still setting errors on graphql engine, do we need to change the docker setting for `:3000` can be connected?
```
graphql-engine_1  | Failed to fetch indexer meta,  request to http://subquery-node:3000/meta failed, reason: connect ECONNREFUSED 172.18.0.3:3000
graphql-engine_1  | Failed to fetch indexer health,  request to http://subquery-node:3000/health failed, reason: connect ECONNREFUSED 172.18.0.3:3000
```

xlc:
Looks like some error in our handler.
It is normal to have some failed connection event initially during starting phase. But it will mean something else is wrong if the connection cannot be restored after few minutes.

shunjizhan:
yup, there are some type mismatch in the handler, already fixed one of them, still investigating the other one. I think they are not related to this PR though, will create another fix CR for the handler

evm-subql/README.md
ThunderDeliverer: Nice catch, fixed 😄

evm-subql/project.yaml
ThunderDeliverer: This persisted from debugging the macOS issues, I reverted it and verified that mac still has the expected behaviour.

ThunderDeliverer:
That error is not due to the PR changes. We will have to address the failing health check issue in another PR. This one allows us to run subql on macOS.

xlc (approved):

PR #99, opened 2021-11-08
implemented getTransactionReceiptByHash

shunjizhan:
## Change
- added config file and command to auto generate types from graphql schema
- added helper function to provide query functions to interact with evm-subql
- implemented getTransactionReceiptByHash function which uses the above helper function to fetch data from evm-subql
- updated README

## Test
- for query helpers, added a couple unit tests
- for eth-rpc-adaptor, tested a couple GET queries to `:8545` and they returned correct result.

### DIFF
--- a/eth-providers/src/base-provider.ts
+++ b/eth-providers/src/base-provider.ts
@@ -25,6 +25,7 @@ import type BN from 'bn.js';
 import { BigNumber, BigNumberish } from 'ethers';
 import {
   BIGNUMBER_ZERO,
+  BIGNUMBER_ONE,
   EFFECTIVE_GAS_PRICE,
   EMPTY_STRING,
   GAS_PRICE,
@@ -40,7 +41,8 @@ import {
   convertNativeToken,
   logger,
   throwNotImplemented,
-  getPartialTransactionReceipt
+  getPartialTransactionReceipt,
+  getTxReceiptByHash
 } from './utils';
 
 export type BlockTag = 'earliest' | 'latest' | 'pending' | string | number;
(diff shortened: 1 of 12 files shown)


eth-providers/src/base-provider.ts
```diff
@@ -741,8 +743,38 @@ export abstract class BaseProvider extends AbstractProvider {
...
+    const tx = await getTxReceiptByHash(transactionHash);
+
+    if (!tx) {
+      return logger.throwError(`transaction hash not found`, Logger.errors.UNKNOWN_ERROR, { transactionHash });
+    }
+
+    // TODO: correct values of these?
+    const confirmations = 0;
+    const effectiveGasPrice = BIGNUMBER_ONE;
+    const byzantium: false = false;
```
xlc, line 756: do we need those?

I am looking at https://infura.io/docs/ethereum/json-rpc/eth-getTransactionReceipt
Also can check https://github.com/paritytech/frontier/blob/master/client/rpc-core/src/types/receipt.rs#L26

https://github.com/paritytech/frontier/blob/4dafbaf5a1a8a0f17f28511f8d9ed100750422b1/client/rpc/src/eth.rs#L1342

eth-providers/src/base-provider.ts
```diff
@@ -741,8 +743,38 @@ export abstract class BaseProvider extends AbstractProvider {
...
+    const defaultAddress = '0x';
+
+    return {
+      to: tx.to || defaultAddress,
+      from: tx.from,
+      contractAddress: tx.contractAddress || defaultAddress,
```
xlc, line 762: according to https://infura.io/docs/ethereum/json-rpc/eth-getTransactionReceipt it should be null if no contractAddress

> the contract address created, if the transaction was a contract creation, otherwise - null.

eth-providers/src/base-provider.ts
```diff
@@ -741,8 +743,38 @@ export abstract class BaseProvider extends AbstractProvider {
...
+    }
+
+    // TODO: correct values of these?
+    const confirmations = 0;
+    const effectiveGasPrice = BIGNUMBER_ONE;
+    const byzantium: false = false;
+    const defaultAddress = '0x';
+
+    return {
+      to: tx.to || defaultAddress,
```
xlc, line 760: according to https://infura.io/docs/ethereum/json-rpc/eth-getTransactionReceipt it should be null if no to

>  address of the receiver. Null when the transaction is a contract creation transaction.

shunjizhan: yeah but this would conflict with the type library we use:
https://github.com/ethers-io/ethers.js/blob/b1458989761c11bf626591706aa4ce98dae2d6a9/packages/abstract-provider/src.ts/index.ts#L106
according to it, `to` and `contractAddress` should both the `string`, also `confirmations`, `effectiveGasPrice`, `byzantium` are required.

Is there a more accurate lib that reflect EVM types? Or should we overwrite `@etherproject` to make the type match infura?

shunjizhan: maybe we should use the parity frontier lib as our type definition?

shunjizhan: The infura definition seems to be the same as [eth wiki](https://eth.wiki/json-rpc/API#eth_gettransactionreceipt), but don't know why `@ethersproject` definition is so wired... Opened a discussion there but no one answered
https://github.com/ethers-io/ethers.js/discussions/2270

xlc: ok. I am ok with the current code until something else complains it needs to be null.
can you write an issue and we can merge this

xlc (approved):

eth-providers/src/base-provider.ts
shunjizhan: sounds good! Created a new issue #100 to address this

Issue #97, opened 2021-11-05
implement `eth_estimateGas`


shunjizhan:

gluneau:
Is it the same as `eth_gasPrice` or different?

shunjizhan:
they are different, `eth-gasPrice` is here #80

PR #101, opened 2021-11-10
Multiple fixes

xlc:
Fix estimate gas, gas limit in block, docker compile file, get transaction receipt

closes https://github.com/AcalaNetwork/bodhi.js/issues/97

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+  "recommendations": ["mrmlnc.vscode-json5"]
+}
--- a/evm-subql/docker-compose.yml
+++ b/evm-subql/docker-compose.yml
@@ -29,22 +29,23 @@ services:
   subquery-node:
     image: onfinality/subql-node:latest
     depends_on:
-      - 'postgres'
+      - postgres
+      - mandala-node
+    ports:
+      - 3000:3000
     restart: always
-    network_mode: host
     environment:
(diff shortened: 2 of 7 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "recommendations": ["mrmlnc.vscode-json5"]
```
shunjizhan, line 2: quick question: should we commit VScode setting files or keep them local?

xlc: this is recommended extension and intended to be committed so others know they should install this.

shunjizhan (approved):

evm-subql/docker-compose.yml
```diff
@@ -29,44 +29,44 @@ services:
...
       DB_DATABASE: postgres
-      DB_HOST: localhost
+      DB_HOST: postgres
       DB_PORT: 5432
       PORT: 3001
     command:
       - --name=acala-evm
       - --playground
-      - --indexer=http://localhost:3000
+      - --indexer=http://subquery-node:3000
```
gluneau, line 69: Does this mean that I have to add these aliases to my `hosts` for running a local testnet?

`127.0.0.1       localhost subquery-node postgres`

xlc: no. docker have its internal dns resolver does which will be used by containers

PR #108, opened 2021-11-12
Rpc endpoint tests

shunjizhan:


shunjizhan:
close it since we have #109, the mocha version

PR #132, opened 2021-11-19
add default env file to use 8545 http port

shunjizhan:
previously in a fresh clone we need to manually add a `.env` file like this every time
```
ENDPOINT_URL=ws://0.0.0.0::9944
HTTP_PORT=8545
```

let't just make it a default to be more convenient


xlc (changes requested):
Instead of modify default env, should just update the default value in code? We want to avoid commit .env file in git

shunjizhan:
added default url and endpoints, so developers won't need to manually add `.env` file every time they clone

xlc (approved):

Issue #133, opened 2021-11-22
`eth_sendRawTransaction` seems to not return the right contract address


gluneau:
Behavior on ganache: Match!
![image](https://user-images.githubusercontent.com/2903505/142950388-4f616801-885b-4621-a415-a21bcfc9c273.png)

Behavior on bodhi: No Match!
![image](https://user-images.githubusercontent.com/2903505/142950499-0fac40c2-55d3-409c-8f72-7763e1bae3f5.png)

Ganache:
```
➜  hardhat npx hardhat run scripts/sample-script.js --network development
Greeter deployed to: 0x3d3593927228553b349767ABa68d4fb1514678CB
➜  hardhat 
➜  hardhat npx hardhat console --network development
Welcome to Node.js v14.18.1.
Type ".help" for more information.
> const Greeter = await ethers.getContractFactory("Greeter");
undefined
> const greeter = await Greeter.attach('0x3d3593927228553b349767ABa68d4fb1514678CB');
undefined
> await greeter.greet()
'Hello, Hardhat!'
> 
(To exit, press Ctrl+C again or Ctrl+D or type .exit)
> 
```

```
Listening on 127.0.0.1:8545
web3_clientVersion
eth_chainId
eth_blockNumber
eth_chainId
eth_chainId
eth_estimateGas
eth_getBlockByNumber
eth_getBlockByNumber
eth_gasPrice
eth_getTransactionCount
eth_chainId
eth_sendRawTransaction

  Transaction: 0xc9c911bfb1b4c9a6ff439a4c012d279dc696fa4120644056a968f8ee91b8a974
  Contract created: 0x3d3593927228553b349767aba68d4fb1514678cb
  Gas usage: 493826
  Block Number: 1
  Block Time: Mon Nov 22 2021 14:42:46 GMT-0500 (GMT-05:00)

eth_chainId
eth_getTransactionByHash
eth_chainId
eth_getTransactionReceipt
web3_clientVersion
eth_chainId
eth_chainId
eth_call
```

Bodhi:
```
➜  hardhat npx hardhat run scripts/sample-script.js --network development
Greeter deployed to: 0xD26e19913ca16B5B59aF7f07472f97cC9eA3f12B
➜  hardhat npx hardhat console --network development                     
Welcome to Node.js v14.18.1.
Type ".help" for more information.
> const Greeter = await ethers.getContractFactory("Greeter");
undefined
[...]

PR #136, opened 2021-11-23
fix nonce

shunjizhan:
## Change
- previously the nonce calculation is not very accurate (usually off by 1), and after hardhat deploys a contract, it uses `address` and `nonce` to calculate target contract address, so won't get correct result. This will fix it.
- changed endpoints of e2e tests to use local host so they can pass locally.
- updated acala contract package version

## Test
- now hello world contract is able to execute function calls.
- `yarn test:dev` still pass in `eth-rpc-adaptor` and `eth-providers`


ntduan (approved):

shunjizhan:
fix #133

PR #141, opened 2021-11-24
fix eip 712

ntduan:


shunjizhan:
in the latest version many of the examples tests will fail, for example when doing `yarn test` in `examples/dex`:
```
 1) Dex
       getLiquidityPool works:
     Error: -32603: execution revert: Other("invalid currency id"): 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001c4f746865722822696e76616c69642063757272656e63792069642229
```

@ntduan

shunjizhan:
ok seems like our mandala docker image is outdated, this works with latest acala node

ntduan:
> **UPDATE:** false alarm
> 
> in the latest version many of the examples tests will fail, for example when doing `yarn test` in `examples/dex`:
> 
> ```
>  1) Dex
>        getLiquidityPool works:
>      Error: -32603: execution revert: Other("invalid currency id"): 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001c4f746865722822696e76616c69642063757272656e63792069642229
> ```
> 
> @ntduan is this error introduced by the eip712 signature so it got unexpected currency id?

This may be caused by a change in the address of the chain contract

PR #146, opened 2021-11-24
support sendTransaction

ntduan:


shunjizhan:
tested around with this branch and most local tests still pass, just one small issue:

Screen Shot 2021-11-24 at 21 09 26

shunjizhan:
I think it might actually be a bug on the subql queries side, I saw subqueries return values are all strings, we can merge this PR, and I will fix the subql types later

shunjizhan (approved):

PR #154, opened 2021-11-30
able to build and start dockerized rpc adaptor

shunjizhan:
## Change
Added Dockerfile to build adn run `eth-rpc-adaptor` service.

Note that this uses `yarn` for itself, instead of `rush` for monorepo, so it needs a slightly different `package.json`, and a `yarn.lock` just for docker (rush won't need to use it)

## Test
- able to start a dockerized `eth-rpc-adaptor` service, and connected to the latest public mandala testnet.
- tested a couple `eth_xxx` and they worked.

## TODO
currently any `eth_xxx` that interacts with evm-subql won't work, we will need to add support for it after we have a public evm-subql service.


xlc (commented):
better to have ./docker/package.json and ./docker/yarn.lock so it is easier to update them and the IDE understand the types

eth-rpc-adapter/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM node:14-alpine
+
+ARG mandala_url=wss://node-6870830370282213376.rz.onfinality.io/ws?apikey=0f273197-e4d5-45e2-b23e-03b015cb7000
```
xlc, line 3: No need to inject this at build time. We can configure environment variable at deployment time.

shunjizhan:
> better to have ./docker/package.json and ./docker/yarn.lock so it is easier to update them and the IDE understand the types

moved

> No need to inject this at build time. We can configure environment variable at deployment time.

removed, now uses local `.env` file, same as running locally

eth-rpc-adapter/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:14-alpine
+
```
tj4laminar, line 2: is it current latest version of node-alpine image ?

eth-rpc-adapter/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ENV NODE_ENV=production
+
```
tj4laminar, line 4: is it required field or just get as it was in example code which shared by Bryan ?

eth-rpc-adapter/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+COPY . .
+COPY docker/package.json ./package.json
+COPY docker/yarn.lock ./yarn.lock
+
+RUN yarn
+RUN [ "yarn", "build" ]
+
+EXPOSE 8545
+
```
tj4laminar, line 15: is it kind of incoming port or just connecting to external world ?

eth-rpc-adapter/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR /usr/src/app
+
+COPY . .
+COPY docker/package.json ./package.json
+COPY docker/yarn.lock ./yarn.lock
+
+RUN yarn
+RUN [ "yarn", "build" ]
+
```
tj4laminar, line 13: let run this container as non-root user

eth-rpc-adapter/Dockerfile
xlc: we want to have this for prod image

eth-rpc-adapter/Dockerfile
xlc: this will be exposed to external world

eth-rpc-adapter/Dockerfile
shunjizhan: latest version is 17 but we use 14 across `bodhi.js` project

eth-rpc-adapter/Dockerfile
shunjizhan: now use user 'node'

tj4laminar (approved):
looks good

PR #160, opened 2021-12-01
fix block error message

ntduan:


eth-providers/src/__tests__/e2e/rpc.test.ts
```diff
@@ -49,4 +49,14 @@ describe('rpc test', () => {
...
     expect(result.toString()).to.equal('0');
   });
+
+  it('getBlockByNumber', async () => {
+    try {
+      const result = await provider._getBlockHeader(
+        '0xff2d5d74f16df09b810225ffd9e1442250914ae6de9459477118d675713c732c'
+      );
+    } catch (error) {
+      expect(error.message === 'header not found');
```
xlc, line 59: best to use expect to throw to ensure the test will fail if it did not throw the expected error

ntduan: Okay, but expect to throw doesn't support asynchronous functions, so I import the package chai-as-promised to implement tests for asynchronous functions.

PR #162, opened 2021-12-01
performance optimization

ntduan:


eth-providers/src/base-provider.ts
```diff
@@ -184,21 +184,22 @@ export abstract class BaseProvider extends AbstractProvider {
...
-          name: this.api.runtimeVersion.specName.toString(),
-          chainId: await this.chainId()
-        };
-
-        this._network = network;
-      } catch (e) {
-        await this.api.disconnect();
-        throw e;
-      }
+      this._network = this.api.isReadyOrError
```
xlc, line 189: this is bit more readable?

```
this._network = (async () => {
try {

         const network = {
           name: this.api.runtimeVersion.specName.toString(),
           chainId: await this.chainId()
         };

          return network;
       } catch (e) {
         await this.api.disconnect();
         throw e;
       }
})()
```

xlc (approved):

PR #168, opened 2021-12-03
Support indexing scheduler evm event

shunjizhan:
## Change
Previously if an evm event is triggered by scheduler, it has no extrinsics, so when trying to get tx hash from extrinsic, it will crash subql. So we added this dummy tx hash for these kind of event.

Note that I chose a very distinguishable dummy tx hash (instead of `0x`), so if this is causing any problem in the future, we will know immediately what is the cause, since the error msg will look something like
```
ERROR: can't find tx hash 0x6666666666666666666666666666666666666666666666666666666666666666
```

## Test
Ran all waffle examples, previously it will crash subql, now it won't. And so far there is no known way to crash subql anymore :tada:

### DIFF
--- a/evm-subql/src/mappings/mappingHandlers.ts
+++ b/evm-subql/src/mappings/mappingHandlers.ts
@@ -6,7 +6,7 @@ import { SubstrateEvent } from '@subql/types';
 import { Log, TransactionReceipt } from '../types';
 
 const NOT_EXIST_TRANSACTION_INDEX = 0xffff;
-const DUMMY_TX_HASH = '0x6666666666666';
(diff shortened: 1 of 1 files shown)


evm-subql/src/mappings/mappingHandlers.ts
```diff
@@ -6,14 +6,16 @@ import { SubstrateEvent } from '@subql/types';
 import { Log, TransactionReceipt } from '../types';
 
 const NOT_EXIST_TRANSACTION_INDEX = 0xffff;
+const DUMMY_TX_HASH = '0x6666666666666';
```
xlc, line 9: can you make it the same length as a standard tx hash

xlc (approved):

PR #170, opened 2021-12-03
rename adaptor to adapter

shunjizhan:


shunjizhan:
is it going to affect the npm package name?

xlc (approved):
it is better to break it now than later

PR #171, opened 2021-12-03
Update dockerfile

xlc:
now the working dir is root dir

`docker build -f eth-rpc-adapter/Dockerfile .`


eth-rpc-adapter/Dockerfile
```diff
@@ -1,18 +1,39 @@
...
 
 COPY . .
-COPY docker/package.json ./package.json
-COPY docker/yarn.lock ./yarn.lock
+RUN rush update
+
+WORKDIR /app/eth-rpc-adapter
+
+RUN rush build -t .
+RUN yarn build
```
shunjizhan, line 22: I think we don't need `yarn build` anymore after `rush build -t .`

tj4laminar:
looks good.. later release , we should reduce its size.

tj4laminar (approved):
👍

Issue #169, opened 2021-12-03
waffle examples failures with public mandala testnet


shunjizhan:
currently all these examples pass with latest local mandala `acala/mandala-node:943f80a5`, but has some failures with the public mandala testenet `wss://node-6870830370282213376.rz.onfinality.io/ws?apikey=0f273197-e4d5-45e2-b23e-03b015cb7000`

shunjizhan:
**dex**
```
10 passing (26s)
  5 failing

1) Dex
       getSwapTargetAmount works:
     Error: -32603: execution revert: Other("Dex get_swap_target_amount failed"): 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002a4f746865722822446578206765745f737761705f7461726765745f616d6f756e74206661696c65642229

 2) Dex
       getSwapSupplyAmount works:
     Error: -32603: execution revert: Other("Dex get_swap_supply_amount failed"): 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002a4f746865722822446578206765745f737761705f737570706c795f616d6f756e74206661696c65642229

  3) Dex
       swapWithExactSupply works:
     Error: -32603: execution revert: Other("InsufficientLiquidity"): 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001e4f746865722822496e73756666696369656e744c69717569646974792229

  4) Dex
       swapWithExactTarget works:
     Error: -32603: execution revert: Other("InsufficientLiquidity"): 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001e4f746865722822496e73756666696369656e744c69717569646974792229

  5) Dex
       addLiquidity and removeLiquidity works:
     Error: -32603: execution revert: Other("InsufficientLiquidity"): 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001e4f746865722822496e73756666696369656e744c69717569646974792229
```

shunjizhan:
## oracle
```
  1 passing (59s)
  1 failing

  1) Prices
       getPrice works:
     AssertionError: Expected "0" to be equal 34500000000000000000000
```

shunjizhan:
## erc20
```
  9 passing (3m)
  3 failing

  1) BasicToken
[...]

PR #172, opened 2021-12-06
able to switch Waffle examples url 

shunjizhan:
## Change
Previously test urls are hardcoded, so it's not super convenient to test with different url. Since now we have a public testnet, it's great to be able to switch between localhost and public url. After these changes, we can do 
```
ENDPOINT_URL=xxx yarn test  # test each
ENDPOINT_URL=xxx ./run test # test all
```

## Test
passing no ENDPOINT_URL, `./run test` still passed locally. And with
```
ENDPOINT_URL=wss://node-6870830370282213376.rz.onfinality.io/ws?apikey=0f273197-e4d5-45e2-b23e-03b015cb7000
```
it will run all tests connecting to the public testnet


zjb0807 (approved):
The public test network may have insufficient balance, so it is better to use a local environment.

shunjizhan:
yeah we did get some `insufficient balance` failure https://github.com/AcalaNetwork/bodhi.js/issues/169, I think in the future it would be great to make all test cases pass not only on local but also on public chian, to make sure the public chain also works as expected.

PR #188, opened 2021-12-15
Docker CI

shunjizhan:
## Change
Finally we have CI to run all these tests for us! **Note that this is only a MVP version**, since @ntduan has some PRs waiting, so we could merge this "usable version" first, and I will do optimazations/refactor in another PR, also please let me know any suggestions for refactor/optimazation.

Currently this will run all e2e tests including:
- `eth-rpc-adaptor` tests
- `eth-provider` tests
- `truffle examples` deploy + test
- `hardhat examples` deploy + test
- `waffle examples` tests

In the future if we imported other contracts as examples, we could also integrate these contracts into CI as well, such as wormhole, linkdrop.

## Test
- all CI tests passed! https://github.com/AcalaNetwork/bodhi.js/actions/runs/1580928038
- if anything goes wrong, CI is expected to catch it for us, so I changed test cases to wrong expectation manually, and CI fails! https://github.com/AcalaNetwork/bodhi.js/actions/runs/1580911334

## TODO
- currently the file structure is random (Dockerfile is in root), maybe should put it in `/test` folder?
- some services need to wait for other independent service to build, need optimazation
- change `COPY --from=bodhi /app /app` to copy only the needed folders
- **let me know your suggestions for refactor/optimazation!**


Dockerfile.bodhi
```diff
@@ -0,0 +1,123 @@
+FROM node:16-alpine as bodhi
```
xlc, line 1: I think if the filename is `bodhi.Dockerfile` then you will have proper syntax highlighting

xlc (approved):

Dockerfile.bodhi
shunjizhan: sounds good, changed. BTW for some reason my vscode can auto detect proper syntax highlighting for `Docker.xxx` as well lol

PR #195, opened 2021-12-16
block full transaction

ntduan:
@xlc  Now transactions contain only evm's transactions


shunjizhan:
some CI tests are failing, looks like another polish we need to have is to print out logs when test failed, not sure why it only show logs when test succeed :thinking_face:

shunjizhan:
there are 2 ways we can reproduce these failure locally:
- follow the [readme](https://github.com/AcalaNetwork/bodhi.js#run-tests-with-docker) and run docker commands, to see logs we can do this in another terminal 
```
docker-compose logs --tail=0 --follow
```
- or we can normally run these tests with locally running service, such as `yarn test:dev`

shunjizhan:
I will double check if these are actual failure, or is it CI's glitch. since this PR passed all CI tests, https://github.com/AcalaNetwork/bodhi.js/pull/196, these might be real failures

shunjizhan:
ran tests locally, and I think the issue is cause by waffle/dex example error
```
  Dex
    1) "before all" hook for "getLiquidityPool works"


  0 passing (1s)
  1 failing

  1) Dex
       "before all" hook for "getLiquidityPool works":
     Error: overflow (fault="overflow", operation="toNumber", value="30166947999407255544874991818928313884966007890769378202518912836159817001281", code=NUMERIC_FAULT, version=bignumber/5.5.0)
```

Since all tests rely on this to "feed TXs" as first step, all tests are failing.

ntduan:
It does fail because I have to wait for a new version to be released before I can update the evm subql code

xlc (approved):

shunjizhan:
> It does fail because I have to wait for a new version to be released before I can update the evm subql code

Sounds good! Is the new version going to have a docker release too?

shunjizhan (approved):
:tada:

PR #217, opened 2022-01-09
Polished Unfinalized Cache and `eth_estimateGas`

### LINKED ISSUE #209: Make `eth_gasPrice()` and `eth_estimateGas()` Compatible
when MetaMask sends native token, it tries to get valid `gasPrice`(Max fee) and `gasLimit` by calling `eth_gasPrice()` and `eth_estimateGas()`, then users should only need to change `tip` (Max priority fee).

However, currently `eth_gasPrice()` is not compatible with `eth_estimateGas()`, and will produce invalid substrate gas params
```ts
// psudo
const gasPrice = eth_gasPrice();
const gasLimit = eth_estimateGas({ gasPrice, ...otherTxParams });
const substrateGasParams = calcSubstrateTransactionParams({
  txGasPrice: gasPrice,
  txGasLimit: gasLimit,
  ...otherConsts,
});   // => bad params
```

### LINKED ISSUE #214: Unfinalized Cache Keep A Couple More Blocks
instead of clearing all finalized block txs, we can keep a sliding window of the latest 10 finalized blocks + unfinalized blocks, so most recent tx can be queries from cache directly. We can even query cache and subql at the same time, and return whichever is faster. This will optimize the performance and stability, will also help many test cases to avoid using database.

shunjizhan:
## Change
- modified `eth_estimateGas` to return gasLimit + storageGasLimit, so after decoding substrate gasLimit won't be negative. fixed #209 
- changed unfinalzied block cache to cache for a couple more blocks, fixed #214 

## Test
- manually tested that metamask can send token directly without having to manually change gasLimit/gasPrice
- modified endpoint tests to use auto generated gasPrice/gasLimit from hard coded gas, and tests still passed
- modified tests for unfinalized cache according to the new behavoir

## Note
These RPCs are intended for metamask auto generated TX to work (such as sending token). However, when deploying a contract, developer still need to calculate gas params himself, since the default storage limit is 640, which is too small to complex TX.

### DIFF
--- a/eth-providers/src/base-provider.ts
+++ b/eth-providers/src/base-provider.ts
@@ -135,6 +135,11 @@ export interface TXReceipt extends partialTX {
   status?: number;
 }
 
+export interface GasConsts {
+  storageDepositPerByte: bigint,
+  txFeePerGas: bigint,
+}
+
 export abstract class BaseProvider extends AbstractProvider {
   readonly _api?: ApiPromise;
   readonly formatter: Formatter;
@@ -539,6 +544,11 @@ export abstract class BaseProvider extends AbstractProvider {
     };
   };
 
+  _getGasConsts = (): GasConsts => ({
+    storageDepositPerByte: (this.api.consts.evm.storageDepositPerByte as UInt).toBigInt(),
+    txFeePerGas: (this.api.consts.evm.txFeePerGas as UInt).toBigInt(),
+  });
+
   /**
    * Estimate gas for a transaction.
    * @param transaction The transaction to estimate the gas of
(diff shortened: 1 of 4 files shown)


eth-providers/src/base-provider.ts
```diff
@@ -546,7 +546,7 @@ export abstract class BaseProvider extends AbstractProvider {
    */
   estimateGas = async (transaction: Deferrable<TransactionRequest>): Promise<BigNumber> => {
     const resources = await this.estimateResources(transaction);
-    return resources.gas;
+    return resources.gas.mul(30); // TODO: this is a temp gas limit, we should optimize gas limit later
```
xlc, line 549: estimateResources provides the storage limit as well, so should use `gas_limit + storage_entry_limit * storage_entry_deposit / tx_fee_per_gas` to return the value

shunjizhan: done

shunjizhan:
merged #220 so close this one

PR #227, opened 2022-01-13
remove EIP-1559 field for getBlock

### LINKED ISSUE #213: Sending Token Consumes High Gas Fee
when sending ACA tokens with metamask, gas fee paid is:
|                        | gasPrice (Gwei) | priority fee (Gwei) | Metamask estimation (ACA) | actual gas cost (ACA) |
|------------------------|-----------------|---------------------|---------------------------|-----------------------|
| with low priority gee  | ~200            | 1                   | 0.03 - 6.8                | 4000                  |
| with high priority gee | ~200            | ~200                | 6.7 - 6.8                 | 20                    |

need to investigate why gas cost is that high, and fix it. Ideally for the best UX, the actual gas cost should be very close the metamask estimation.

shunjizhan:
## Change
As discussed in #213 , we will make Metamask use Legacy format to send token, and it will do it if it doesn't see EIP-1559 fields in block data.

fix #213 

## Test
- Previously we already had test for metamask sending token using legacy format, they still pass
- manually did some testing and sending tokens still work

### DIFF
--- a/eth-providers/src/base-provider.ts
+++ b/eth-providers/src/base-provider.ts
@@ -354,7 +354,7 @@ export abstract class BaseProvider extends AbstractProvider {
       author: author,
       extraData: EMPTY_STRING,
 
-      baseFeePerGas: BIGNUMBER_ZERO,
+      // baseFeePerGas: BIGNUMBER_ZERO,
       transactions
     };
 


eth-providers/src/base-provider.ts
```diff
@@ -354,7 +354,7 @@ export abstract class BaseProvider extends AbstractProvider {
       author: author,
       extraData: EMPTY_STRING,
 
-      baseFeePerGas: BIGNUMBER_ZERO,
+      // baseFeePerGas: BIGNUMBER_ZERO,
```
xlc, line 357: should add some comments explain this

ntduan (approved):

PR #250, opened 2022-01-25
implement `eth_subscribe`

shunjizhan:
## Change
implemented `eth_subscribe` method, currently it only support `newHeads` and `logs` params.

## Test
- unit tested log filter function
- manually tested subscribe and unsubscribe, and they can send back logs/heads info from websocket

### DIFF
--- a/eth-providers/src/base-provider.ts
+++ b/eth-providers/src/base-provider.ts
@@ -178,21 +178,17 @@ export abstract class BaseProvider extends AbstractProvider {
     await this.isReady();
 
     this.api.rpc.chain.subscribeNewHeads(async (header: Header) => {
+      // cache
       const blockNumber = header.number.toNumber();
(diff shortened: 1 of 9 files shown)


eth-providers/src/base-provider.ts
```diff
@@ -169,6 +188,48 @@ export abstract class BaseProvider extends AbstractProvider {
     this.api.rpc.chain.subscribeFinalizedHeads(async (header: Header) => {
       this._cache!.handleFinalizedBlock(header.number.toNumber());
     }) as unknown as void;
+
+    // TODO: factor this out or maybe rename it to start subscription?
+    this.api.rpc.chain.subscribeNewHeads(async (header: Header) => {
+      if (this._listeners[NEW_HEADS]?.length > 0) {
```
xlc, line 194: why not merge this into the other `subscribeNewHeads`? we will want to avoid multiple subscriptions to reduce load on the node side

PR #261, opened 2022-02-01
add getEthResources

ntduan:


eth-providers/src/base-provider.ts
```diff
@@ -650,6 +651,79 @@ export abstract class BaseProvider extends AbstractProvider {
     return resources.gas.add(storageGasLimit);
   };
 
+  /**
+   * Get the gas for eth transactions
+   * @returns The gas used by eth transaction
+   */
+  getEthResources = async (
+    transaction: Deferrable<TransactionRequest>,
```
shunjizhan, line 659: we can make `transaction` param optional, so then we will be able to use this method directly to calculate eth gasLimit/gasPrice when we don't know the whole transactions

shunjizhan: something like
```
const ethGasParams = getEthResources({
  transactions: null,
  gasLimit,
  gasPrice,
  storageLimit,
})
```

ntduan: If the transaction parameter is not known, then a function can be designed separately.

PR #265, opened 2022-02-09
getFeeData polish

shunjizhan:
## Change
removed eip-1559 fields from getFeeData, this will make our provider compatible with wormhole relayer

## Test
previously there will be some gas calculation issue when used with relayer, now there isn't


xlc (approved):
at later stage we should have a cli flag to toggle between legacy mode vs eip1559 mode

PR #267, opened 2022-02-10
clean up and fix publish script

shunjizhan:
## Change
- fixed publish script to work with submodule
- did some clean up to remove useless files

@xlc @tj4laminar just wanna double check `eth-rpc-adapter/docker` is not used by anything, right? I remember it was for some very early testing, so I removed it


xlc:
Yeah I think should be fine.

Issue #274, opened 2022-02-16
Configuration to track how many best blocks to store in the in memory store


xlc:
Currently we save last 200 blocks in the in memory store.
Should make this configurable and potentially make it infinite, which makes local development easier by avoiding the subql services.

Should print some big warning if infinite is used.

shunjizhan:
closed by #280

PR #275, opened 2022-02-17
Safe Mode and Configurable Cache Size

### LINKED ISSUE #273: Configuration to only track finalized heads
Should have a configuration to enable safe mode that only return finalized data.

Should add extra safeness check before return any block hash / transaction hash to make sure the transaction / block are finalized.

shunjizhan:
## Change
Added two parameters:
- `SAFE_MODE` to enable safe mode
- `MAX_CACHE_SIZE` to specify max cached blocks

Will print warning when safe mode is on, or cache size too big.

fix #273 #274 

## Details
Since all unfinalized info is managed by subscribing to latest heads, we have an [3-line](https://github.com/AcalaNetwork/bodhi.js/compare/safe-mode?expand=1#diff-8e05d01e1b336d3501a54605915fd1c9116c51bc23417352c54359c4ca197119R213-R217) elegant solution for safe mode, which makes it only subscribe to finalized head when safe mode is on. As a result:
- cache behaves like subql (only cache finalized blocks), so only finalized tx can be found.
- subscription callbacks are only triggered by new finalized heads, so only finalized logs will be notified.

## Test
It will take ~50 hours to test it with code, so I just manually tested:
- cache only cached `MAX_CACHE_SIZE` number of blocks
- when safe mode enabled, `eth_subscribe` only returns logs from finalized blocks, and a transaction hash can only be found after it's block is finalized.
- when safe mode disabled, `eth_subscribe` returns logs from latest blocks, and a transaction hash can be found as soon as it's in latest block.


xlc (approved):
environment variable is fine but later we should support configuration via cli args and have a `--help` to print the help text

xlc:
also I am not sure if subscribe to finalized head is enough.
e.g if people query a unfinalized block directly, it should be treated as block not found, instead of return the unfinalized block.
basically we want a helper to check if a tx hash / block hash is finalized, and every time we want to return a block / tx, it needs to ensure the data is finalized or treated as non existent.

eth-providers/src/base-provider.ts
```diff
@@ -1233,6 +1233,21 @@ export abstract class BaseProvider extends AbstractProvider {
...
   };
 
+  _isBlockFinalized = async (blockTag: BlockTag): boolean => {
+    const [finalizedHead, verifyingBlockHash] = await Promise.all([
+      this.api.rpc.chain.getFinalizedHead(),
+      this._getBlockHash(blockTag)
+    ]);
+
+    const [finalizedBlockNumber, verifyingBlockNumber] = (
+      await Promise.all([this.api.rpc.chain.getHeader(finalizedHead), this.api.rpc.chain.getHeader(verifyingBlockHash)])
```
xlc, line 1243: this is fine but in future we could cache some of those calls

shunjizhan: sounds good, we can make it work first and do some optimazation in the future

shunjizhan:
closes this for #280

eth-providers/src/base-provider.ts
shunjizhan: I think the current performance is not too bad, already parallellized these fetches as much as possible. In the future we can do some more caching to make it O(1)

PR #280, opened 2022-02-18
Safe Mode and Configurable Cache Size (WIP)

### LINKED ISSUE #273: Configuration to only track finalized heads

shunjizhan:
this is a rebased and updated version of #275 

## Change
Added two parameters:
- `SAFE_MODE` to enable safe mode, so unfinalized blocks will behave like not exist
- `MAX_CACHE_SIZE` to specify max cached blocks

Will print warning when safe mode is on, or cache size too big.

fix #273 #274 

### for querying TX and log subscription
Since all unfinalized caching info is managed by subscribing to latest heads, we can make subscription only subscribes to finalized head when safe mode is on. As a result:
- cache behaves like subql (only cache finalized blocks), so only finalized tx can be found.
- subscription callbacks are only triggered by new finalized heads, so only finalized logs will be notified.

### for other RPCs that can pass blockTag as parameters
Added a check to make sure this target block is finalized, if not, throws an unfinalized error, this should make the RPC call behaves the same as block not found.

## Test
manually tested:

### max cache size
- cache only cached `MAX_CACHE_SIZE` number of blocks

### for querying TX and log subscription
**scope**
- eth_getTransactionByHash
- eth_getTransactionReceipt
- eth_subscribe

**result**
- when safe mode enabled, `eth_subscribe` only returns logs from finalized blocks, and a transaction hash can only be found after it's block is finalized.
- when safe mode disabled, `eth_subscribe` returns logs from latest blocks, and a transaction hash can be found as soon as it's in latest block.

### for other RPCs that can pass blockTag as parameters
**scope**
- eth_call
- eth_getTransactionCount
- eth_getCode
- eth_getBalance
- eth_getStorageAt
- eth_getTransactionByBlockHashAndIndex
- eth_getTransactionByBlockNumberAndIndex
- eth_getBlockTransactionCountByHash
- eth_getBlockTransactionCountByNumber

**result**
|             | finalized block       | unfinalized block                                       | non-exist block                               |
|-------------|-----------------------|---------------------------------------------------------|-----------------------------------------------|
| safe mode   | return correct result | return `null` or throw "target block not finalized" error | return `null` or throw "header not found" error |
| normal mode | return correct result | return correct result                                   | return `null `or throw "header not found" error |


common/git-hooks/pre-commit
```diff
@@ -6,4 +6,4 @@
 # Invoke the "rush prettier" custom command to reformat files whenever they
 # are committed. The command is defined in common/config/rush/command-line.json
 # and uses the "rush-prettier" autoinstaller.
-node common/scripts/install-run-rush.js prettier || exit $?
\ No newline at end of file
+# node common/scripts/install-run-rush.js prettier || exit $?
```
xlc, line 9: do you intent to disable this?

eth-providers/src/rpc-provider.ts
```diff
@@ -3,15 +3,25 @@ import { BaseProvider, BlockTag } from './base-provider';
 import { createApi } from './chain-api';
 import { throwNotImplemented } from './utils';
 
+export interface BaseProviderOptions {
+  safemode?: boolean;
```
xlc, line 7:
```suggestion
  safeMode?: boolean;
```

common/git-hooks/pre-commit
shunjizhan: yeah this is giving some wired auto fix, which makes the style worse lol, so not really need it

eth-providers/src/rpc-provider.ts
shunjizhan: resolved

xlc (approved):

PR #306, opened 2022-03-14
getEthGas polish

shunjizhan:
### previous interface
```
{
  "method": "eth_getEthGas",
  "params": [21000000, 64100, 1000000]
}
```

### new interface
```
{
  "method": "eth_getEthGas",
  "params": [{
      "gasLimit?": 21000000,
      "storageLimit?": 64100,
      "validUntil?": 1000000
  }]
}
```
There are two advantages:
- named params are easier for developers to follow
- now every param is optional, so developers can do the following default query to get default gas params for big gas operations, such as contract deployment.  
```
{
  "method": "eth_getEthGas",
  "params": []
}
----- OR -----
{
  "method": "eth_getEthGas",
  "params": [{}]
}
```
which returns same result as
```
{
  "method": "eth_getEthGas",
  "params": [{
      "gasLimit": 21000000,
      "storageLimit": 64100,
      "validUntil": currentBlock + 150
  }]
}
```

## Test
added tests accordingly

### DIFF
--- a/eth-providers/src/base-provider.ts
+++ b/eth-providers/src/base-provider.ts
@@ -191,7 +191,6 @@ export abstract class BaseProvider extends AbstractProvider {
     this.safeMode = safeMode;
     this.latestFinalizedBlockHash = undefined;
 
-    logger.info(`safe mode: ${safeMode}`);
     safeMode && logger.warn(`
       ----------------------------- WARNING ----------------------------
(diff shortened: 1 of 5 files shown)


shunjizhan:
hold on one sec there is one small type

shunjizhan:
@gluneau @ThunderDeliverer could you double check the new interface makes sense? If it does I will merge it and deploy to public mandala, thanks

PR #311, opened 2022-03-20
`eth_getTxReceipt` polish

### LINKED ISSUE #290: optimize getTXByHash
currently we will retry 20 times, because we need to wait for next block to be able to find it, in this case we will need to wait 20 seconds if tx is not found.

Can think about ways to optimize this, maybe we can cache those "sent but not mined TX" in rpc-adaptor

shunjizhan:
## Change
improved how `eth_getTxReceipt` works. 

### for TX that exists
Previously it keeps asking tx in a loop, and there were two issues:
- if tx doesn't go into block for 20s, it will fail to find it. 
- if there are many request coming in, and each of them run in a loop, it's huge performance burden.

Now we will first check if the tx is in the pending state, and if it is, we will wait until it's included in the block by a listener. So there is only 1 query instead of 20. This also solves the issue when TX isn't included in the block in time, it can still wait until target tx is in the block and return it.

### for TX that doesn't exists
Previously it still need to query 20 times, and return not found after 20s.

Now it can return not found in 3s.

fix #290 

## Test
- all previous tests still pass
- manually tested that:
  - querying a pending tx works as expected
  - querying a tx that's included in the cache works as expected


zjb0807 (commented):
```
    "message": "Error: transaction hash not found (txHash=\"0xec290bd8c36361fa8db1f674ef561abec2e767a5fca557ab6c13f420bd14eb7f\", code=UNKNOWN_ERROR, version=bodhi.js/providers/2.3.1)"
```
I think return null is better

zjb0807 (approved):

shunjizhan:
sounds good I will do this change in next PR

PR #313, opened 2022-03-22
Fix receipt and warning

zjb0807:


eth-providers/src/base-provider.ts
```diff
@@ -1007,7 +1007,7 @@ export abstract class BaseProvider extends AbstractProvider {
       storageLimit = BigInt(_storageLimit);
       validUntil = BigInt(_validUntil);
       tip = BigInt(_tip);
-    } else if (ethTx.type == null || ethTx.type === 0 || ethTx.type === 2) {
+    } else if (ethTx.type === null || ethTx.type === 0 || ethTx.type === 2) {
```
shunjizhan, line 1010: `ethTx.type == null` is the same as `ethTx.type === undefined || ethTx.type === null`

shunjizhan (approved):

Issue #125, opened 2021-11-16
Implement eth_getTransactionByHash with pending tag


ntduan:

zjb0807:
https://github.com/AcalaNetwork/bodhi.js/pull/316#discussion_r832869766

zjb0807:
Closes by #323

PR #316, opened 2022-03-23
`eth_getTransactionReceipt` polish

shunjizhan:
## Change
`eth_getTransactionReceipt` returns `null` instead of throwing error when tx not found.

## Test
changed tests accordingly and all tests passed

### DIFF
--- a/eth-providers/src/base-provider.ts
+++ b/eth-providers/src/base-provider.ts
@@ -1600,6 +1600,7 @@ export abstract class BaseProvider extends AbstractProvider {
 
   _getTXReceipt = async (txHash: string): Promise<TransactionReceipt | TransactionReceiptGQL | null> => {
     // TODO: potentially these 3 can go in parallel
+    let res = null;
     try {
       if (await this._isTXPending(txHash)) {
         const txFromNextBlock = await this._getTXReceiptFromNextBlock(txHash);
@@ -1610,18 +1611,21 @@ export abstract class BaseProvider extends AbstractProvider {
       if (txFromCache) return txFromCache;
 
       const txFromSubql = await this.subql?.getTxReceiptByHash(txHash);
-      return txFromSubql || null;
+      res = txFromSubql || null;
     } catch {
-      return null;
+      res = null;
     }
+
+    return res;
   };
 
   // Queries
   getTransaction = (txHash: string): Promise<TransactionResponse> =>
     throwNotImplemented('getTransaction (deprecated: please use getTransactionByHash)');
 
-  getTransactionByHash = async (txHash: string): Promise<TX> => {
+  getTransactionByHash = async (txHash: string): Promise<TX | null> => {
     const tx = await this._getTXReceipt(txHash);
+    if (!tx) return null;
 
     const extrinsic = await this._getExtrinsicsAtBlock(tx.blockHash, txHash);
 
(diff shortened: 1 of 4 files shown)


eth-rpc-adapter/src/eip1193-bridge.ts
```diff
@@ -311,8 +294,8 @@ class Eip1193BridgeImpl {
   async eth_getTransactionByHash(params: any[]): Promise<TX> {
```
zjb0807, line 294:
```suggestion
  async eth_getTransactionByHash(params: any[]): Promise<TX | null> {
```

eth-providers/src/base-provider.ts
```diff
@@ -1572,29 +1571,34 @@ export abstract class BaseProvider extends AbstractProvider {
...
   };
 
   // Queries
   getTransaction = (txHash: string): Promise<TransactionResponse> =>
     throwNotImplemented('getTransaction (deprecated: please use getTransactionByHash)');
 
-  getTransactionByHash = async (txHash: string): Promise<TX> => {
+  getTransactionByHash = async (txHash: string): Promise<TX | null> => {
     const tx = await this._getTXReceipt(txHash);
+    if (!tx) return null;
```
zjb0807, line 1601: The tx maybe is pending, need to return the tx(some fields null).

shunjizhan: https://eth.wiki/json-rpc/API

> Returns
> Object - A transaction receipt object, or null when no receipt was found

shunjizhan: if tx is pending, we will wait for it to be included in the block, then return the full receipt (with a timeout), this has been our behavior since the very beginning. 

I am fine if we want to change the behavior, but this will be a big breaking change and should be discussed more by the team

shunjizhan: Oh I found an issue from long ago, we can keep track of it https://github.com/AcalaNetwork/bodhi.js/issues/125

shunjizhan: as far as I remember, `eth_getTransactionReceipt` is very heavily used by all of the tools, such as metamask, hardaht, truffle, etc.. but `eth_getTransactionByHash` is rarely used, so I think #125 is not urgent

zjb0807: `eth_getTransactionByHash` have no `pending` tag, only one param.  I think we can do it now.

eth-providers/src/base-provider.ts
```diff
@@ -1572,29 +1571,34 @@ export abstract class BaseProvider extends AbstractProvider {
...
 
     const extrinsic = await this._getExtrinsicsAtBlock(tx.blockHash, txHash);
```
zjb0807, line 1604:
```
const from = tx.from ||  await this.getEvmAddress(args.origin).toLowerCase();// not sure

return {
      from: from,
      to: args.target || null,
      hash: tx.transactionHash,
      blockHash: tx.blockHash || null,
      nonce,
      blockNumber: tx.blockNumber || null,
      transactionIndex: tx.transactionIndex || null,
      value,
      gasPrice: GAS_PRICE,
      gas: args.gas_limit,
      input
    };
```

zjb0807 (approved):

eth-providers/src/base-provider.ts
shunjizhan:
```
const tx = await this._getTXReceipt(txHash);
```
if tx is pending, this will call `_getTXReceiptFromNextBlock()`, so we will need some more refactor to make #125 work. So let's keep this PR in it's scope to fix `eth_getTransactionReceipt`

shunjizhan:
I think we can add a flag to `_getTXReceipt` to change how it deals with pending tx, either wait for next block, or return null directly

PR #317, opened 2022-03-23
Fix block size and gas gasPrice

zjb0807:
Need to update evm-subql


eth-providers/src/base-provider.ts
```diff
@@ -551,6 +482,120 @@ export abstract class BaseProvider extends AbstractProvider {
     return data;
   };
 
+  _parseExtrinsic = async (
```
shunjizhan, line 485: As far as I remember `_getBlock` doesn't have any test, so in the future we can extract this big transformer to a pure helper, and add some tests to it, but for now as long as we manually know it works it's OK

shunjizhan (approved):

eth-providers/src/base-provider.ts
zjb0807: We should use  `expect().to.deep.equal({})` to test the returned fields and encoding. 
This is not very easy to implement. we can use `--instant-sealing` flag to generate mock data.

PR #323, opened 2022-03-25
Standardized Contract Deployment/Call Process

shunjizhan:
## Context
A long time ago, when we were able to deploy hello world for the first time, due to some too-old-unreproducable-but-not-important-anymore issue, we had this workaround that waits until the tx is mined to return tx receipt, for at most 20s.
#126  

However, this is not a very standard workflow, the **standard flow** for pending tx should be:
- `eth_getTxByHash` should return partial tx data, and some fields being `null`
- `eth_getTxReceipt` should return `null`

## Breaking Changes
Changed `eth_getTxByHash` and `eth_getTxReceipt`'s behavior to the standard way. 

Now both of the RPCs should return **immediately** without waiting for tx to be mined, and client itself should handle when tx is mined. For examples:
- Metamask will keep asking `eth_getTxReceipt` until it's not null after sending token
- hardhat will keep asking `eth_getTxReceipt` until it's not null after deploying a contract. However, hardhat doesn't wait automatically for contract calls, so if we want hardhat to work with normal-sealing, we need to manually add `tx.wait()`.

As a result, we will need a couple polish for hardhat tutorials (partial [PR](https://github.com/AcalaNetwork/hardhat-tutorials/pull/31)):
- we don't need to loop helper anymore!
- in order for contract call to work with normal-sealing node, we need to add `tx.wait()` for all state changing calls.

This PR also polished `getTransactionCount` to be able to get more accurate nonce by adding the pending nonce.

### normal-sealing
manually tested:
- truffle tutorials all passed
- hardhat Token tutorials passed after adding `tx.wait()`
- metamask token sending still works 

### instant-sealing
after removing loop helper, all previous CI tests still passed with instant sealing

### RPC behavior
- https://eth.wiki/json-rpc/API
- https://stackoverflow.com/a/67263711/5770676

### Hardhat behavior
- https://discord.com/channels/750408878008827925/750408878008827928/954325283597582356
Screen Shot 2022-03-28 at 15 33 47


eth-providers/src/base-provider.ts
```diff
@@ -1606,13 +1632,27 @@ export abstract class BaseProvider extends AbstractProvider {
...
+      blockHash: null,
+      blockNumber: null,
+      transactionIndex: null,
+      hash: txHash,
+      nonce: targetExtrinsic.nonce.toNumber(),
+      value: args.value,
+      gasPrice: args.storage_limit,   // TODO: reverse calculate if needed
+      gas: args.gas_limit,
+      input: args.input,
+    }
```
zjb0807, line 1655: missed
```
      v: DUMMY_V,
      r: DUMMY_R,
      s: DUMMY_S,
```

zjb0807: And check `extrinsic.method.section.toUpperCase() === "EVM"`

shunjizhan: that's a good point, I will add it tomorrow.

zjb0807 (approved):
LGTM

shunjizhan:
Thanks @zjb0807 , I will merge it and if it's incompatible with latest mandala I will upload fix in another PR

PR #334, opened 2022-04-01
Storage query performance optimization

ntduan:


shunjizhan:
question: why we don't cache `blockHash => Storage`, and instead cache `registry => Storage`? It seems that we are repeating this deterministic query
```
const registry = await this.api.getBlockRegistry(u8aToU8a(blockHash));
```

shunjizhan:
question: why we don't cache `blockHash => Storage`, and instead cache `registry => Storage`? It seems that we are repeating this deterministic query
```
const registry = await this.api.getBlockRegistry(u8aToU8a(blockHash));
```

ntduan:
> question: why don't we cache `blockHash => Storage`, and instead cache `registry => Storage`? It seems that we are repeating this deterministic query (is this line actually not deterministic?)
> 
> ```
> const registry = await this.api.getBlockRegistry(u8aToU8a(blockHash));
> ```
> 
> Also, do we have a limit for the cache? It can grow very big in long term

Because the registry only changes when the chain version is upgraded. This will not be very frequent.

apiAt creates a new object every time, which can cause a significant performance overhead. https://github.com/polkadot-js/api/blob/master/packages/api/src/base/Init.ts#L120

PR #337, opened 2022-04-01
add local mode

shunjizhan:
## Change
Added a local mode for testing with `instant-sealing` local node with hardhat. These 2 RPCs behaves slightly different than usual to aid hardhat behavior.

## Test
WIP


shunjizhan:
@ThunderDeliverer if you use this branch and run a local mode, should now be able to see all hardhat tests passed without `tx.wait()`, no matter with `instant-sealing` or normal sealing local node . Also for AAVE tests, won't need to add `tx.wait()` either. Please let me know if you find is any issue, thanks!

Issue #342, opened 2022-04-04
return the error message instead of error index


zjb0807:
```
{"level":20,"time":1648824912201,"pid":28,"hostname":"7673b738547b","id":3966362797572,"jsonrpc":"2.0","method":"eth_estimateGas","params":[{"from":"0xeb4ffd6eaf0baccd5616b9e4042bd117bc03169a","value":"0xfe7e0cde8321023","gasPrice":"0x2eadc4000a","data":null,"to":"0x896c12e689665283530bd1ece3c29af55e23bb46"}],"msg":"incoming request"}
[...]

PR #347, opened 2022-04-07
Return error message instead of index

zjb0807:
Closes: #342

### DIFF
--- a/eth-rpc-adapter/src/router.ts
+++ b/eth-rpc-adapter/src/router.ts
@@ -4,7 +4,7 @@ import { InvalidParams, JSONRPCError, MethodNotFound } from './errors';
 import { logger } from './logger';
 import { JSONRPCResponse } from './transports/types';
 
-const ErrorRegex = /execution fatal: Module { index: (\d+), error: (\d+), message: None }/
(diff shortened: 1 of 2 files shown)


eth-rpc-adapter/src/router.ts
```diff
@@ -4,6 +4,8 @@ import { InvalidParams, JSONRPCError, MethodNotFound } from './errors';
 import { logger } from './logger';
 import { JSONRPCResponse } from './transports/types';
 
+const ErrorRegex = /execution fatal: Module { index: (\d+), error: (\d+), message: None }/
```
shunjizhan, line 7:
```suggestion
const errorRegex = /execution fatal: Module { index: (\d+), error: (\d+), message: None }/
```

eth-rpc-adapter/src/router.ts
```diff
@@ -40,7 +42,14 @@ export class Router {
 
       logger.error({ err, methodName, params }, 'request error');
 
-      return { error: { code: 6969, message: `Error: ${err.message}` } };
+      let message = err.message;
+      const match = message.match(ErrorRegex);
```
shunjizhan, line 46:
```suggestion
      const match = message.match(errorRegex);
```

shunjizhan (approved):

PR #352, opened 2022-04-10
Remove formatter

zjb0807:
https://github.com/ethers-io/ethers.js/blob/ef1b28e958b50cea7ff44da43b3c5ff054e4b483/packages/providers/src.ts/formatter.ts#L232-L236

https://github.com/ethers-io/ethers.js/blob/ef1b28e958b50cea7ff44da43b3c5ff054e4b483/packages/address/src.ts/index.ts#L89

The `formatter` will convert the address to the format of the checksum, which will fail when the string matches


shunjizhan (approved):

ntduan:
I don't really recommend removing the formatter directly. base-providers were originally intended to be compatible with ethers.js . If you just want to return the rpc result without checksum, just change the hexlifyRpcResult function. Like this: 

image

PR #365, opened 2022-04-13
uodate subql

### LINKED ISSUE #302: Upgrade Subql to Latest Version
currently we are using an old version of subql, and there have been a major upgrade, so might be great to keep it to latest

ntduan:
Fixes #302


shunjizhan (approved):
do we not need to change the config and it worked directly? That's quite a surprise

zjb0807:
> do we not need to change the config and it worked directly? 

It works. Upgrade `project.yaml` later https://github.com/subquery/subql/issues/895

PR #367, opened 2022-04-14
add cache prune and polish health check

shunjizhan:
## Change
Simplified cache logic to auto shrink it's size to `maxCacheBlocks`

Also polished the health check to handle RPC err and timeouts

## Test
added test to `getHealthResult` and modified cache test accordingly

### DIFF
--- a/eth-providers/src/utils/BlockCache.ts
+++ b/eth-providers/src/utils/BlockCache.ts
@@ -7,7 +7,7 @@ export interface BlockToHashesMap {
 }
 
 export interface CacheInspect {
-  extraBlockCount: number;
+  maxCachedBlocks: number;
   cachedBlocksCount: number;
   cachedBlocks: string[];
   allBlockToHash: Record<string, string[]>;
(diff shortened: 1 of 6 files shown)


zjb0807 (commented):
https://github.com/isaacs/node-lru-cache
LRU implementation should be better

shunjizhan:
> https://github.com/isaacs/node-lru-cache LRU implementation should be better

I think LRU is more useful for the storage cache https://github.com/AcalaNetwork/bodhi.js/blob/master/eth-providers/src/base-provider.ts#L214

For block cache, a sliding window of fixed size should be ok

zjb0807:
> For block cache, a sliding window of fixed size should be ok, since we know (defined) that the block cache targets the queries for latest blocks (especially unfinalized onces), no matter how often they are called

Using LRU, we can no longer care about and maintain the size of the cache, just add and get

shunjizhan:
yeah so I agree that's a great use case for the storage cache https://github.com/AcalaNetwork/bodhi.js/blob/master/eth-providers/src/base-provider.ts#L214, which doesn't have a size limit for now, if we use LRU for it, it can maintain a size automatically.

For block cache, the use case and definition is different than LRU. We don't want to maintain "most queried X blocks", but explicitly want "latest X blocks". For examples: if there are X latest unfinalized blocks, we don't care how many times they are queries, we still need to cache them. If a old block is queries 10000 times, we still don't have to cache it, but handle it to subql

eth-providers/src/utils/BlockCache.ts
```diff
@@ -7,44 +7,51 @@ export interface BlockToHashesMap {
...
   }
 
-  handleFinalizedBlock(blockNumber: number): void {
-    const blockToRemove = blockNumber - this.extraBlockCount;
-    this.blockTxHashes[blockToRemove]?.forEach((h) => delete this.allTxHashes[h]);
-    delete this.blockTxHashes[blockToRemove];
+  // if block exist in cache, remove it, otherwise do nothing
+  _removeBlock(blockToRemove: number): void {
+    this.blockToHashes[blockToRemove]?.forEach((h) => delete this.hashToBlocks[h]);
+    delete this.blockToHashes[blockToRemove];
```
zjb0807, line 43: check `undefined` before deleting.

shunjizhan: check which one `undefined`?

shunjizhan: `this.blockToHashes` is always defined

zjb0807: `this.blockToHashes`  may be not consecutive numbers, such as the `ws` connect interrupt

shunjizhan: deleting empty stuff is ok
```
const x = {}
delete x.aaaaa.  // => true
```

shunjizhan: please see my comment here https://github.com/AcalaNetwork/bodhi.js/pull/367/files#diff-505de5c3567ec34cbf3e2ba1a24e5356e13a445a3a53a18c152f89419af660e5R40

shunjizhan: also from here it should already made sure target is defined
https://github.com/AcalaNetwork/bodhi.js/pull/367/files#diff-505de5c3567ec34cbf3e2ba1a24e5356e13a445a3a53a18c152f89419af660e5R35-R36

zjb0807: LGTM

zjb0807 (approved):

PR #380, opened 2022-04-17
Use @acala-network/type-definitions

ntduan:


shunjizhan:
@xlc switching to `@acala-network/type-definitions` solved all the deps warning, I am wondering do we still need to switch `~` to `^`? If yes, should we do this to all packages, or only `@polkadot` packages?

PR #385, opened 2022-04-18
fix waffle package to avoid vulnerability

shunjizhan:
## Change
updated pacakges  to avoid vulnerability for `@openzeppelin/contracts < 4.4.2`

## Test
all waffle examples still pass


common/config/rush/common-versions.json
```diff
@@ -71,10 +71,6 @@
       "^4.2.0",
       "^4.0.2",
     ],
-    "@openzeppelin/contracts": [
-      "~3.4.2",
-      "~4.3.2"
```
zjb0807, line 76: 3.4.2 for `solidity ^0.6.0`
4.4.2 for `solidity ^0.8.0`
both are required here

shunjizhan:
after some digging, arbitrager has to use the older version of openzeppelin lib, otherwise it won't be compatible with the underlying lib https://github.com/Uniswap/v2-periphery/blob/master/contracts/UniswapV2Migrator.sol#L1

PR #389, opened 2022-04-20
upgrade subql config

shunjizhan:
## Change
upgraded subql config yml to latest v1.0.0, also updated all types to `types.json`, which solves the issue that it didn't work with Karura.

## Test
- current tests still pass
- previously there is a `Unable to decode storage system.events::...` error on Karura at block 55,795, was able to reproduce locally, and after this upgrade, it's able to correctly index that block locally.


evm-subql/project.yaml
```diff
@@ -1,467 +1,32 @@
...
+  chaintypes:
+    # TODO: can generate this dynamically with `yarn build` using `@acala-network/types`
+    file: ./types.json
 dataSources:
-  - name: main
-    kind: substrate/Runtime
+  - kind: substrate/Runtime
     startBlock: 1
     mapping:
+      file: ./dist/index.js
```
zjb0807, line 29: This path is error

zjb0807: We need lists:
```
project-mandala.yaml
project-karura.yaml (startBlock: 1)
project-karura-1780000.yaml (startBlock: 1780000)
project-acala.yaml (startBlock: 1)
project-acala-840000.yaml (startBlock: 840000)
project-default.yaml (0.0.1 version)
```

shunjizhan: what error? after `yarn build` this path exist

zjb0807: Ok. And update the `"@acala-network/eth-providers": "~2.4.6"`

evm-subql/project-mandala.yaml
```diff
@@ -0,0 +1,30 @@
...
+    name: '@subql/query'
+    version: 0.14.1
+description: 'subquery for Public Mandala EVM+'
+repository: 'https://github.com/AcalaNetwork/bodhi.js/tree/master/evm-subql'
+schema:
+  file: ./schema.graphql
+network:
+  # public mandala
+  chainId: '0x5c562e6300954998233c9a40b6b86f3028977e6d32d0da1af207738d19f98c1b'
+  endpoint: wss://karura-rpc-0.aca-api.network
```
zjb0807, line 18:
```suggestion
  endpoint: wss://mandala-tc7-rpcnode.aca-dev.network/ws
```

zjb0807 (commented):
Not sure if `project-dev.yaml (0.0.1 version)` is required for CI

shunjizhan:
> Not sure if project-dev.yaml (0.0.1 version) is required for CI

I don't think so, Ci can uses the latest v1.0.0 and still and passed

zjb0807:
> I don't think so, CI is already using the latest `project.yml` (v1.0.0) and still passed
We delete the `project.yml` default file and not use `-f` flag . How does it work?

shunjizhan:
we didn't delete the default `project.yml`, it's there

shunjizhan:
click `Load Diff` you will see

zjb0807 (approved):
LGTM

PR #398, opened 2022-04-26
Revert "remove fixed deps"

shunjizhan:
## Change
Reverts AcalaNetwork/bodhi.js#390

waffle examples will still have the deps warning without these locks :
- with lock, no warnings: https://github.com/AcalaNetwork/bodhi.js/runs/6102942438?check_suite_focus=true
- without lock, has warnings: https://github.com/AcalaNetwork/bodhi.js/runs/6104611253?check_suite_focus=true

This is because waffle examples rely on `@acala-network/api` which has [outdated dependecies](https://github.com/AcalaNetwork/acala.js/blob/master/packages/api/package.json#L22), so this warning should only affect waffle examples. I think it should be fine if we just leave it as is, and after `@acala-network/api` is updated the errors will go awao. 

What do you think @ntduan ? Do you prefer keep the locks or remove them

### DIFF
--- a/common/config/rush/.pnpmfile.cjs
+++ b/common/config/rush/.pnpmfile.cjs
@@ -18,32 +18,31 @@ module.exports = {
   }
 };
 
-// const polkadotDeps = {
-//   "@polkadot/api": "~8.1.1",
-//   "@polkadot/api-derive": "~8.1.1",
-//   '@polkadot/api-augment': "~8.1.1",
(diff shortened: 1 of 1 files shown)


ntduan:
I checked that the waffle examples, doesn't actually use @acala-network/api, so the dependencies can be removed.  Removing polkadot dependencies in the lock file can help us catch dependency problems. Until we can't solve the dependency problem, at that time we can use lock.

PR #403, opened 2022-04-29
add TECHNICALCOMMITTEE

zjb0807:
```
curl --location --request POST 'https://eth-rpc-karura.aca-api.network' \
--data-raw '{
    "jsonrpc": "2.0",
    "method": "eth_getBlockByNumber",
    "params": [
        1818395, true
    ],
    "id": 1
}'
```


zjb0807:
@xlc Can we handle all scenes except EVM in default? Or add it when it appears.

shunjizhan (approved):

PR #418, opened 2022-05-03
eth-rpc-adapter DataDog basic custom instrumentation

HussainMehdi:
Basic datadog custom instrumentation added to monitor methods elapsed time
Set `EXTENSIVE_DD_INSTRUMENTATION=true` in env to enable instrumentation

### DIFF
--- a/eth-rpc-adapter/src/transports/server-transport.ts
+++ b/eth-rpc-adapter/src/transports/server-transport.ts
@@ -44,7 +44,7 @@ export abstract class ServerTransport {
             method,
             params
           },
-          enterTime: Date.now(),
+          enterTime: performance.now(),
           exitTime: -1,
           elapsedTime: -1
         }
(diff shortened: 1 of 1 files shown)


eth-rpc-adapter/src/transports/server-transport.ts
```diff
@@ -34,6 +35,20 @@ export abstract class ServerTransport {
...
+    // Get datadog span from the context
+    const span = process.env.EXTENSIVE_DD_INSTRUMENTATION == 'true' ? tracer.scope().active() : null;
+    // Initialize datadog span tags
+    const spanTags = span
+      ? {
+          body: {
+            method,
+            params
+          },
+          enterTime: Date.now(),
```
xlc, line 47: should use `performance.now()` for benchmarking purpose

HussainMehdi: Fixed!!
Thanks 👍

xlc (approved):

PR #427, opened 2022-05-11
get ChainId from EvmAccount

zjb0807:


shunjizhan (approved):

bodhi/src/Signer.ts
```diff
@@ -286,7 +286,7 @@ export class Signer extends Abstractsigner implements TypedDataSigner {
                 gasPrice: BigNumber.from(1),
                 data: dataToString(data),
                 value: BigNumber.from(tx.value || '0'),
-                chainId: +this.provider.api.consts.evm.chainId.toString(),
+                chainId: +this.provider.api.consts.evmAccounts.chainId.toString(),
```
xlc, line 289: better to be able to detect which one is available to be able to handle both new and old runtime

zjb0807: The chainId of evmAccounts always exists, so it can be compatible with new and old runtime

PR #435, opened 2022-05-13
[net_health] don't fetch very old random block

HussainMehdi:
Problem:  `net_health` fetches random block to instrument time, due to which very old blocks are mostly fetched and that takes longer than fetching recent finalized blocks. Since `ApiPromise` is shared so it slows down other rpc calls of `WsProvider`

Solution: Added Limit to fetch random block but not below `BLOCK_DISTANCE`

_FUTURE TODO_ **Permenant Solution**: Should have multiple load balanced `ApiPromise` in `base-provider.ts`

### DIFF
--- a/eth-providers/src/base-provider.ts
+++ b/eth-providers/src/base-provider.ts
@@ -1901,8 +1901,10 @@ export abstract class BaseProvider extends AbstractProvider {
       })
     );
 
+    // Distance allowed to fetch old random block (since most oldest block takes longer to fetch)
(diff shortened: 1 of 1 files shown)


shunjizhan (approved):

xlc:
what's the reason of using random block? not just head - 100?

HussainMehdi:
> what's the reason of using random block? not just head - 100?

Yes can do that, since it was random was already there so didn't bothered changing that. 

Updated!!

xlc (approved):

PR #461, opened 2022-06-01
add id in JSONRPCResponse

zjb0807:
https://www.jsonrpc.org/specification

```
id
This member is REQUIRED.
It MUST be the same as the value of the id member in the Request Object.
If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.
```

### DIFF
--- a/eth-rpc-adapter/src/transports/server-transport.ts
+++ b/eth-rpc-adapter/src/transports/server-transport.ts
@@ -20,7 +20,7 @@ export abstract class ServerTransport {
   }
 
   protected async routerHandler({ id, method, params }: JSONRPCRequest, cb?: any): Promise<JSONRPCResponse> {
-    if (!id || !method) {
+    if (id === null || id === undefined || !method) {
       console.error(`invalid json request: id: ${id}, method: ${method}, params: ${params}`);
       const error = new InvalidRequest();
       return {


eth-rpc-adapter/src/transports/server-transport.ts
```diff
@@ -20,9 +20,11 @@ export abstract class ServerTransport {
   }
 
   protected async routerHandler({ id, method, params }: JSONRPCRequest, cb?: any): Promise<JSONRPCResponse> {
-    if (id === null || id === undefined || !method) {
+    if (!id || !method) {
```
shunjizhan, line 23:
```suggestion
    if (id === null || id === undefined || !method) {
```

shunjizhan: `!id` will be true if id is `0` or `''`, so usually we still want `id === null || id === undefined` to be more safe

eth-rpc-adapter/src/transports/http.ts
```diff
@@ -59,6 +59,7 @@ export default class HTTPServerTransport extends ServerTransport {
       if (req.body.length > this.options.batch_size) {
         const error = new InvalidRequest();
         result = {
+          id: null,
```
shunjizhan, line 62:
```suggestion
          id: req.body?.id ?? null,
```

> If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.

from my understanding id should only be `null` when there is no id in the request body （fail to detect）, but I think either way should be fine since I checked infura eth rpc response they don't even have an id for invalid json request……

shunjizhan (approved):

shunjizhan:
Screen Shot 2022-06-01 at 16 24 51
oh wait they do sometimes have id for error

eth-rpc-adapter/src/transports/http.ts
zjb0807: the body is an array, each item has an id, so if I don't know which one, I use the `null`.

```
[
    {
        "id": 16,
        "jsonrpc": "2.0",
        "method": "eth_getBalance",
        "params": [
            "0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1",
            "latest"
        ]
    },
    {
        "id": 226,
        "jsonrpc": "2.0",
        "method": "eth_getBalance",
        "params": [
            "0x7dda0dc1c85bfe07f959fb6ed59be8486f48afc5",
            "0xB0EE4"
        ]
    }
]
```

PR #466, opened 2022-06-08
core refactors

### LINKED ISSUE #336: test and polish for `_getBlock`
currently the largest chunk of untested code is the `getBlock` part, this already started to cause some issue (new code can build and pass all tests, but it still broke something). we should add some tests to it, otherwise any new code related to it will be very dangerous.

There are a couple things we can do:
- e2e test for `eth_getBlockByHash` for `full = true` and` full = false`, and make sure returned data is correct
- polish `_parseExtrinsic`, add TS to it, and maybe factor it out as a pure function helper, add test to it
- polish `_getBlock` and changed all `any` to real types

shunjizhan:
## Context
Previously we iteratively implement each functionality without knowing how much of them have overlaps, so basically for each core function it's a gigantic code block, and they have a lot of overlapping logics. But since they are too big, it's hard to reuse the logics, so we should decompose them into smaller pieces, and make the code more DRY.

Sorry for the super big PR, but there is really no intermediate stage I can land on, it's like a "everything works" or "everything breaks" situation. 

### refactors
- refactors to break down big chunks to smaller and reusable pieces, and they are reused as much as possible:
  - get all info that can the parsed from extrinsic alone with `parseExtrinsic`
  - get transaction info with a helper `_parseTxAtBlock`
  - calculate gas with `getEffectiveGasPrice`
- some refactor for `getblock`, especially simplified how to calculate full block
- previously we had a couple AbstractProvider methods overrides with force wrong type, this might cause issues if other code uses them, since the return types are wrong. After the refactor we avoid using these, and defined our own method names.
- linter fixes: remove unused vars, add func return type, etc... now the `rush build` has much less warnings

### new stuff
- able to get `effectiveGasPrice` in tx receipt
- able to index effectiveGasPrice into subql (currently still have some small issue so not in action yet)

fix #336 

## Test
- all previous tests passed, including the [super strict ones](https://github.com/AcalaNetwork/bodhi.js/commit/bdda05c29472a2ccc991664a206f8003b1470339#diff-97c55078074d22eab24ac3c6a198deba9758e7826f43e355b7edea803fc220cd), which should be able to enforce that the responses from the couple core RPCs (`eth_getBlock*`, `eth_getTxByHash`, `eth_getTxReceipt`) are the same as before, plus the new `effectiveGasPrice` field.

## Todo
- still need to find a workaround for subql to be able to query prev block to calculate correct gasPrice
- then restart mandala with this version, and index from the beginning
- publish a new pacakge after this PR is merged, so subql might be able to reuse the above helpers, we will see

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -84,8 +84,8 @@ services:
     depends_on:
       - eth-rpc-adapter-server-with-subql
       - eth-rpc-adapter-server-public-mandala
-    environment:
-      - SKIP_PUBLIC=1
+    # environment:
+    #   - SKIP_PUBLIC=1
 
   # ------------------------------------ #
   # ---------- No Subql Realm ---------- #
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -84,6 +84,8 @@ services:
     depends_on:
       - eth-rpc-adapter-server-with-subql
       - eth-rpc-adapter-server-public-mandala
+    environment:
+      - SKIP_PUBLIC=1
 
   # ------------------------------------ #
   # ---------- No Subql Realm ---------- #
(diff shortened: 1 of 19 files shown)


shunjizhan:
I am gonna merge this and if there is any changes needed I will do it in separate commits

PR #476, opened 2022-06-16
Add forward mode

### LINKED ISSUE #474: Forward mode
Add new cli to enable forward mode that forwards Substrate RPCs to Substrate node.

Use `rpc_methods` to query all the supported RPC methods from Substrate node and only forward those.

Should be disabled by default.

ntduan:
close #474


shunjizhan:
can we also add the forward mode description in https://github.com/AcalaNetwork/bodhi.js/tree/master/eth-rpc-adapter#options? thanks

shunjizhan (approved):
looks good to me as long as the tests passed either locally or after re-run

PR #479, opened 2022-06-17
fix to in `eth_getTransactionByHash`

shunjizhan:
## Change
fixed `to` in `getTxByHash`, for non-evm extrinsics, we can't parse `to` from extrinsic. Fortunately we already parsed `to` from txReceipt, so we can use it directly

## Test
- added tests to make sure `getTxByHash` returns correct result for dex.swap extrinsic


shunjizhan:
so for `dex.swapWithExactSupply([currencyIdA, currencyIdB], ...)`, `to` returns `TokenA.address`, not sure if this makes sense though. We can keep it this way for now

zjb0807 (approved):

PR #482, opened 2022-06-20
Subql block handler

shunjizhan:
## Change
- subql now supports indexing `effectiveGasPrice`, and receipts from dex.swap transactions for erc20s
- subql now use block handler instead of event handler, there are two reasons:
  - since we need to make api calls for each event, by handling block we can fire api calls parallel.
  -  for dex.swap transactions, we will need to combine multiple events, so can't handle them separately

## Test
- added new e2e tests that made sure subql correctly indexed the dex.swap transactions.
- manually checked that getting receipt from subql is the same as getting receipt from cache

## Todo
haven't tested what happens for dex.swap between erc20 and native token.


evm-subql/src/mappings/mappingHandlers.ts
```diff
@@ -88,24 +84,31 @@ export async function handleEvmEvent(event: SubstrateEvent): Promise<void> {
...
 
     await log.save();
   }
 }
+
+export async function handleBlock(block: SubstrateBlock): Promise<void> {
+  const TARGET_MODULES = ['EVM', 'DEX'];
+  const targetExtrinsics = block.block.extrinsics.filter((e) =>
+    TARGET_MODULES.includes(e.method.section.toUpperCase())
+  );
```
zjb0807, line 111: We still need to filter the events of evm module. `honzon-bridge`/`currencies`/`load` and other modules will include the evm tx.

zjb0807: use `findEvmEvent` to filter the EVM tx

shunjizhan:
@zjb0807 now it should includes all modules, and we did a filter for these with evm events so to avoid too many extra xhrs

PR #507, opened 2022-07-11
Add timestamp to logs

eliasmpw:


shunjizhan:
do we also need timestamp for transaction receipt? Since we already added it to log, I am thinking if we should also do it for transaction receipt in case later we need it

shunjizhan:
if not we can merge this

eliasmpw:
I just checked and I will not need timestamps on transaction receipt right now, but I agree that it could be really useful later, so I'm adding it

shunjizhan (approved):

PR #511, opened 2022-07-18
Rich Mode

shunjizhan:
## Change
added a rich mode for more convenient local contract testing. In this mode gas params will default to large amount so all operations should not require at gas overrides anymore.

This is especially helpful for large contract tests with hardhat.

## Test
Tested some local contract deployments without gas overrides:
- when rich mode is enabled, contract creation succeed
- when rich mode disabled, contract creation will fail by `Error: evm.OutOfStorage: Storage usage exceeds storage limit`


zjb0807 (approved):
It works for me

PR #518, opened 2022-07-26
Update mandala node

zjb0807:


shunjizhan:
I will remove the gasPrice part in the test since it changes for every version

shunjizhan:
I removed the gasPrice part, we can add it back when doing some big chunk of code editing that might break gas, but normally won't need them

PR #530, opened 2022-08-03
add support for orphan events and virtual tx

shunjizhan:
## Change
added support for these event transactions that doesn't have tx hash with it, let's call them orphan events. We wrap these events in a "virtual tx", so we can still see them and query them. 

We can't assign each of the log a new virtual tx, but should group them together, so related logs from the same event are contained in one virtual tx.

We define the tx hash as `keccak256(blockHash + eventIndex)`

## Test
manually (test script) tested all the following standard for virtual tx / orphan logs 
✔️ `eth_subscribe` can notify the logs triggered by scheduler
✔️ `eth_getLogs` can find the orphan logs
✔️ `eth_getTransactionReceipt` can find the virtual tx receipt
✔️  `eth_getTransactionByHash` can find the virtual tx
✔️  `eth_getBlockByxxx` can find the virtual tx receipt
✔️ virtual tx receipt and orphan logs are indexed in subquery
✔️ virtual tx hash can be found in cache


eth-providers/src/base-provider.ts
```diff
@@ -1772,6 +1775,28 @@ export abstract class BaseProvider extends AbstractProvider {
...
+    const allEvents = await this.queryStorage<Vec<FrameSystemEventRecord>>('system.events', [], blockHash);
+
+    const evmLogs = allEvents
+      .filter(isPhantomEvmEvent)
+      .map(getPartialTransactionReceipt)
+      .map((receipt, i) =>
+        receipt.logs.map((log) => ({
+          ...log,
+          transactionIndex: 0,
+          transactionHash: keccak256(blockHash + i + i),
```
xlc, line 1790: why `+ i` twice? also this the hash of the string right? I think we want to hash the raw data rather hex string

xlc (commented):
Not sure about the name of PhantomEvmEvent because there are very real. Maybe orphan evm event? Events without parents.

shunjizhan:
@xlc I have finished local testing and looks good, so it's ready for some CR again, so once wormhole confirmed they are able to re-observe the VAA we can merge it ASAP and let guardians deploy new version.

xlc (approved):
Some tests to ensure virtual hash are calculated as expected will be good

PR #545, opened 2022-08-18
update subql doc

shunjizhan:


evm-subql/README.md
```diff
@@ -113,5 +127,33 @@ query {
...
+}
+```
+## For Production
+Previous examples are **local dev setup**, which uses the [example config](./project.yaml) that is tailored to local dev node. 
+
+For production, we need slightly different configs, and usually need to start each of the `{ node, postgres, indexer, query }` seperately with docker or k8s, instead of running them all in one docker container.
+
+One trick is that we don't have to stat indexing from block 0, since Acala and Karura didn't enable evm until some certain block. In particular we can use these two configs for prod (change `endpoint` to your custom one if needed):
+- [Acala prod](./project-acala-840000.yaml)
+- [Karura prod](./project-acala-840000.yaml)
```
xlc, line 152: is this right?

shunjizhan:
it's fixed https://github.com/AcalaNetwork/bodhi.js/commits/master

PR #547, opened 2022-08-19
Update subql-node.Dockerfile

ayudh37:
Was getting an [error](https://stackoverflow.com/questions/54428608/docker-node-alpine-image-build-fails-on-node-gyp) when doing `docker-compose up` for the subquery-node service so added that line.

I'm using an Apple M1 Machine

### DIFF
--- a/docker/subql-node.Dockerfile
+++ b/docker/subql-node.Dockerfile
@@ -3,5 +3,6 @@ FROM onfinality/subql-node:v1.7.0 as subql-node
 COPY evm-subql /app/evm-subql
 
 WORKDIR /app/evm-subql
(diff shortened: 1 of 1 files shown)


shunjizhan:
@ThunderDeliverer you are using M1 chips, right? Can you confirm this line is needed to run

ThunderDeliverer:
That's right, M1 Max has no issues with running the Dockerfile. It just show the standard errors until the first block is mined.

@ayudh37  do you have Rosetta installed?

ntduan:
I think you can try to upgrade the docker version and see if the issue is solved.

PR #556, opened 2022-09-09
Narrow ethers imports

stwiname:
Theres a couple of reasons for this change:
- Greatly reduces the code size when bundling 
- Drops the imports of networking modules (http) so that it can be run in the SubQuery sandbox.


shunjizhan:
thanks!

PR #559, opened 2022-09-21
update rpc test script

shunjizhan:


.github/workflows/test.yml
```diff
@@ -23,62 +23,64 @@ jobs:
...
+        run: docker volume rm bodhijs-postgres || true && docker rm -vf $(docker ps -a | grep bodhijs-subquery-node | awk '{print $1}') || true && docker rm -vf $(docker ps -a | grep bodhijs-graphql-engine | awk '{print $1}') || true && docker rm -vf $(docker ps -a | grep bodhijs-postgres | awk '{print $1}') || true && docker rm -vf $(docker ps -a | grep bodhijs-mandala-node | awk '{print $1}') || true && docker rm -vf $(docker ps -a | grep bodhijs-eth-rpc-adapter-server | awk '{print $1}') || true  && docker rm -vf $(docker ps -a | grep bodhijs-eth-rpc-adapter-server-with-subql | awk '{print $1}') || true  && docker rm -vf $(docker ps -a | grep bodhijs-feed-tx | awk '{print $1}') || true && docker rm -vf $(docker ps -a | grep bodhijs-eth-rpc-adapter-test | awk '{print $1}') || true
```
xlc, line 55: `docker compose down -v` should clean everything

shunjizhan: it removes everything, is it an issue? I remember a while ago we wanted to make sure it doesn't accidentally clear other processes

shunjizhan: oh so it only clears containers and volumes related to the compose file, so we should be good

PR #561, opened 2022-09-26
unified http and ws ports

### LINKED ISSUE #557: eth-rpc-adapter: support unified port for both http & ws
allow the http port and ws port to be the same one to make everything easier

shunjizhan:
## Change
Now it can accept both ws and http requests on the same port.

There are also a couple cleanups and polishes:
- combined `HttpServerTransport` and `WsServerTransport` to one `EthRpcServer` to make logics more concise.
- removed forward mode which we don't need
- removed some http2 related logic that we don't seem to use
- fix the `eth_unsubscribe` return data when subscription does not exist

fix #557 

## Test
- old tests still pass
- previously we didn't have any ws related tests (only manually tested), so I took the chance to add some ws tests to make sure ws subscription works, which can also ensure future changes won't break ws part

## Todo
- clean up & polish options with #555
- #563


eth-rpc-adapter/src/index.ts
```diff
@@ -1,5 +1,62 @@
...
+  }
+
+  console.log(`
+  --------------------------------------------
+               🚀 SERVER STARTED 🚀
+  --------------------------------------------
+  version         : ${version}
+  endpoint url    : ${opts.endpoints}
+  subquery url    : ${opts.subqlUrl}
+  listening to    : http ${opts.httpPort} | ws ${opts.wsPort}
```
ii-ii-ii, line 48: Should we remove the wsport option?

eth-rpc-adapter/src/server.ts
```diff
@@ -1,77 +1,227 @@
...
+    app.use(cors(corsOptions));
+    app.use(jsonParser({ limit: '1mb' }));
+    app.use(this.httpHandler.bind(this));
+    app.use(errorHandler);
 
-  const bridge = new Eip1193Bridge(provider);
+    this.server = http.createServer(app);
 
-  const rpcForward = opts.forwardMode ? new RpcForward(provider) : undefined;
+    this.wss = new WebSocket.Server({ server: this.server });
```
ntduan, line 60: Confirmed with bryan that it would be nice to have a switch to control whether to start the websocket server or not.

shunjizhan: sounds good, I can add it in a separate commit to prevent this PR to get over complicated. By the way what's the use case for only allowing http but not ws port?

ntduan: I think it's based on performance considerations, and it doesn't really matter.

ntduan (approved):

eth-rpc-adapter/src/server.ts
shunjizhan: cool, I can add it along with #555

PR #580, opened 2022-10-21
signer and waffle examples flow polish

shunjizhan:
### flow update
polished the testing flow, previously the setup is a little bit too circle
```
test file => utils.getTestProvider() => new TestProvider() => getWallets()
```
it also requires an additional util file for each of the examples. 

now all the testing setup is encapsulated into a helper function from bodhi directly  
```
test file => getTestUtils()
```
so ppl can understand more easily by checking `getTestUtils`'s code, which is consistent with the updated example in the README

### other polishes
- some style fix, such as removed unused variables, removed unnecessary ts-ignore, etc. 
- some minor code polish, renamed `AccountSigningKey` to `PolkaSigner` which I think is more descriptive, to distinguish it from evm signer.
- updated docs and examples for bodhi and waffle-examples

## Test
all waffle examples are changed accordingly, they still works as expected


examples/waffle/e2e/test/e2e.test.ts
```diff
@@ -1,46 +0,0 @@
-import { expect, use } from 'chai';
```
shunjizhan, line 1: looks like a trivial test, so removed it

bodhi/src/evmChai/supportEmit.ts
```diff
@@ -78,7 +78,9 @@ export function supportEmit(Assertion: Chai.AssertionStatic): void {
       try {
         assertArgsArraysEqual(context, expectedArgs, log);
         return;
-      } catch (_a) {}
+      } catch {
+        // ignore
```
shunjizhan, line 82: this line will somehow suppress the ts `empty block` warning lol

examples/waffle/utils.ts
```diff
@@ -1,51 +0,0 @@
-import { TestProvider, AccountSigningKey, Provider, Signer } from '@acala-network/bodhi';
```
shunjizhan, line 1: in old setup, although we had this shared util file, it's actually a little bit hacky. If you look at it's position, it's not suppose to work, because it doesn't belong to any package, and shouldn't find the `node_modules`. 

It only happen to work in rush context where `node_modules` in flattened in the root.

PR #584, opened 2022-10-26
Update example

zjb0807:


docker-compose-example.yml
```diff
@@ -78,12 +78,11 @@ services:
...
-      - 3331:3331
     extra_hosts:
       - host.docker.internal:host-gateway
     environment:
       - SUBQL_URL=http://graphql-engine:3001
       - ENDPOINT_URL=ws://mandala-node:9944
       - LOCAL_MODE=0
       - SAFE_MODE=0
-    command: yarn start
+    command: yarn start -l=0 -s=0
```
shunjizhan, line 88:
```suggestion
    command: yarn start
```
it should be the same after https://github.com/AcalaNetwork/bodhi.js/pull/585

shunjizhan: -l=0 -s=0 is the default options

shunjizhan (approved):

PR #594, opened 2022-11-24
update deps

shunjizhan:
## Change
- updated `@polkadot/api` from `8.1.1` to latest `9.9.1`, also updated sub deps for it
- updated `ethers` from to `5.4.0` to latest `5.7.0`
- updated github action to use latest versions
- simplified the creation process of providers
- some type fixes according to the new auto-generated types

close #593 

## Test
all tests still pass, except one, can fix it later


shunjizhan:
@zjb0807 do you have any idea why [this test](https://github.com/AcalaNetwork/bodhi.js/pull/594/files#diff-a6b14ead2b2a67990851fdc60642518627f441529c7b39b2fbcb02e0e8ced616R57-R60) is failing? It kept returning 0, did we ever change the default behavior for this (from returning max u128 to 0)?

PR #596, opened 2022-11-30
replace `rpc.evm.call` with `state_call`

shunjizhan:
## Change
use `state_call` to implement a eth call, similar to how the rust version did. Now the eth calls won't invoke any `api.rpc.evm.call` anymore.

## Test
- added some unit tests for the helper functions
- manually tested a couple simple `eth_call` and worked
- many of the existing e2e tests call `eth_call` directly or indirectly, which still pass, so I didn't add more tests for `eth_call` itself.

### DIFF
--- a/eth-providers/package.json
+++ b/eth-providers/package.json
@@ -15,8 +15,8 @@
     "gql:typegen": "graphql-codegen --config codegen.yml"
   },
   "dependencies": {
-    "@acala-network/api": "~4.1.8-2.2",
-    "@acala-network/types": "~4.1.8-2.2",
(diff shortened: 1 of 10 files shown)


eth-providers/package.json
```diff
@@ -41,6 +41,8 @@
     "ethers": "~5.7.0",
     "graphql": "~16.0.1",
     "graphql-request": "~3.6.1",
+    "bn.js": "~5.2.0",
+    "@types/bn.js": "~5.1.0",
```
xlc, line 45: should be in `devDependencies`

xlc (approved):

PR #633, opened 2023-01-20
support subql logs compare

shunjizhan:
## Change
- now subql compare can compare both tx receipts and logs, which is all we have in evm subql. So now it's able to do a full diff of subql database
- polished compare method to speed up
- support writing output to file

## Test
added tests for logs compare


common/git-hooks/pre-commit
```diff
@@ -7,4 +7,4 @@
 # are committed. The command is defined in common/config/rush/command-line.json
 # and uses the "rush-prettier" autoinstaller.
 # node common/scripts/install-run-rush.js prettier || exit $?
-node common/scripts/install-run-rush.js prettier || exit $?
+# node common/scripts/install-run-rush.js prettier || exit $?
```
xlc, line 10: why change this?

scripts/subql-diff/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+### run with example data
+first extract the data, which contains some valid karura receipts/logs, and another copy of data that missed some record.
+```
+tar -xvf data.tgz
+```
+
+check for receipts diff
+```
+yarn diff                                \
```
xlc, line 13: better to have this published to npm and use `npx` to run so people don't need to clone & setup this repo

common/git-hooks/pre-commit
shunjizhan: I feel like prettier is too opinionated and inflexible which actually uglifies many codes...

shunjizhan: eslint is much better

xlc (approved):

PR #642, opened 2023-02-01
update deps

ntduan:
This might avoid confusion about the version of the polkadot/api when others use eth-provider as a dependency. But it also brings a problem that the user has to install polkadot/api himself when using

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -125,11 +125,7 @@ importers:
       '@ethersproject/providers': 5.7.2
       '@ethersproject/transactions': 5.7.0
       '@ethersproject/wallet': 5.7.0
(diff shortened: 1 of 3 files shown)


xlc (approved):

shunjizhan:
I am not super familiar with the best practices of peer dependencies, just wondering why some `@polkadot` deps are dependencies (such as `keyring`), some of them are peer dependencies?

ntduan:
> I am not super familiar with the best practices of peer dependencies, just curious why some `@polkadot` deps are dependencies (such as `keyring`), some of them are peer dependencies, such as `api`?

If they are placed in peerDependencies, then the user must install both the api and the keyring.

zjb0807 (approved):

shunjizhan (approved):

shunjizhan:
close it as it's outdated

PR #643, opened 2023-02-01
added util to get all receipts at block

### LINKED ISSUE #635: Regression tests
Some regression tests to ensure the latest version still works with historical blocks to prevent bugs like https://github.com/AcalaNetwork/bodhi.js/pull/634

shunjizhan:
## Context
When we first build up bodhi, we didn't have an overview at that moment of all sorts of transactions we will need to support. As a result, our approach was **bottom up**: when we know a txHash, we look for it's details. And when we know many txHashes, we look for each of their details, one by one.

### Drawbacks
This approach had a couple drawbacks: 
- **inefficient**: for every tx, we need to go through all extrinsics and events in a block in order to find it's details, and makes a lot of duplicated async calls. This not only affect rpc adapter's performance, but also increase node's burden.
- **lacks encapsulation**: to get a receipt, the call path involves many helpers inside and outside the base provider. These methods' dependencies are deep and circular, so it's hard to debug or update. 
- **poor compatibility with subql**: 
  - subql's block mapping handler already provide much block data. However, we went very circular by extracting txHashes first, and query for their info again.
  - since our receipt logic is not encapsulated enough, as mentioned above, we often need to copy and paste changes from bodhi to subql
- **hard to test**: 
  - we usually won't notice a subql bug until an indexer failed in prod.
  - most receipt tests are e2e, which requires setting up the whole stack including rpc adapter, so testing is very inconvenient.

### Solution
In order to solve these issues, we will switch to a **top down** approach: find and cache all tx receipts from a block. When we look for a tx receipt, we don't need to make a chain of async calls anymore, it should be either from cache or subql. This approach has many merits:
- **efficient**: to find all `N` txs in each block, only 1-pass is needed, which is much more efficient than previous `N` pass. This can also dedupe a lot of async calls, by ensuring each async call is called only once. We can then upgrade cache to save actual receipts, instead of only txHash, so querying cache will become instant.
- **great encapsulation**: all receipts related logic is encapsulated into a top `getAllReceiptsAtBlock` method, with it's sub logics as sub-methods. It's logic graph is tree-like, so each time we need to debug/modify/add logic to receipts, we can easily start from the root, and search for the sub-methods.
- **perfectly compatible with subql**: to index each block, subql won't need to have it's own codes anymore. It should be able to call `parseReceiptsFromBlockData` directly, and do some minimal format convertion. 
- **easy to test**: now we don't need to setup anything for testing. We can directly test the standalone `getAllReceiptsAtBlock` util against known edge case blocks. When a new edge case come out, we can simply add it to the test suite as regression test. Fix #635 

## Change
Added the essential `getAllReceiptsAtBlock` util. In order to keep each PR to readable size, that's all. Next PR(s) will actually plug this into bodhi/subql, and eliminate legacy stuff.

## Test
added unit tests against some normal blocks, and a couple known edge case blocks. All results matched current receipts
  - send kar
  - deploy contract
  - contract call
  - orphan tx
  - erc20 xcm
  - new/legacy gas info format
  

## Todo
1) these unit tests won't be 100% comprehensive, so after merging we will update subquery and index all data (in dev), then compare with old ones to see if there is any issue
2) after confirming all receipts in subquery has no issue, plug this util into bodhi and refactor cache
3) improve the logic to get receipt using the new cache


eth-providers/src/__tests__/e2e/parseBlock.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    it('with some other random txs in the block', async () => {
+      const blockNumber = 2669090;
+      const receipts = await getAllReceiptsAtBlockNumber(apiA, blockNumber);
+
+      expect(receipts.length).to.equal(1);
+      expect(receipts[0]).to.deep.equal(acala2669090);
+    });
+
+    it.skip('multiple xcms', async () => {
+      // didn't find any block like this yet...
```
zjb0807, line 181: https://acala.subscan.io/event?block=2832519

PR #645, opened 2023-02-03
new subql handler

shunjizhan:
## Change
some magic happened that reduced mapping handler from 200 lines to 40 lines

a couple small improvement and fixes:
- previously normal tx's id had some small issue that it always started with 2, now it's back to 0
- previously orphan tx's id is `${blocknumber}-${somehash}`, now it's consistent with normal tx: `${blockNumber}-${txIndex}`

## Test
- prev tests still pass
- will deploy a dev version and see

### DIFF
--- a/evm-subql/.eslintrc.js
+++ b/evm-subql/.eslintrc.js
@@ -1,0 +1,45 @@
+module.exports = {
+  env: {
+    browser: false,
+    es2021: true,
+  },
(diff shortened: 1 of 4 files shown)


shunjizhan:
I am gonna merge it directly to deploy a dev subql, but do let me know if you have any comments

PR #655, opened 2023-02-08
block cache v2

shunjizhan:
## Change
after #643, we can now easily cache all receipts in a block. This new version of block cache keeps track of a sliding window of blocks, and stores all receipts in that range.

in the future, subql + blockCache should be the only interface for finding receipt, i.e., to fetch a receipt, we won't need to make any extra async calls  anymore, we only need to ask cache **synchronously**, or ask subql with **1 async call**. This is much more efficient than the previous approach, which usually need **a chain of** async calls to get a receipt.

## Test
added unit tests for all methods of the new block cache

## Todo
Note that the scope of this PR is limited to implementing the new block cache, and some minimum modification to plug it into base-provider. Next PR(s) will actually utilize this more powerful cache to optimize all logics.


eth-providers/src/utils/BlockCache.ts
```diff
@@ -1,57 +1,54 @@
...
+export type BlockNumToReceipts = Record<number, FullReceipt[]>;
 
 export interface CacheInspect {
   maxCachedBlocks: number;
   cachedBlocksCount: number;
-  cachedBlocks: string[];
-  allBlockToHash: Record<string, string[]>;
-  allHashToBlock: Record<string, number>;
+  hashToReceipt: HashToReceipt;
+  blockNumToReceipts: BlockNumToReceipts;
```
zjb0807, line 10: `blockNumber` is unreliable, there is a fork, we should use `blockHash`

eth-providers/src/utils/BlockCache.ts
```diff
@@ -1,57 +1,59 @@
-export interface HashToBlockMap {
-  [hash: string]: number;
-}
+import { FullReceipt } from './transactionReceiptHelper';
 
-export interface BlockToHashesMap {
-  [block: string]: string[];
-}
+export type HashToReceipt = Record<string, FullReceipt>;
```
zjb0807, line 3:
```suggestion
export type TxHashToReceipt = Record<string, FullReceipt>;
```

zjb0807 (approved):

PR #656, opened 2023-02-10
new receipt flow

shunjizhan:
### new receipt flow
bodhi new internal flow

### other updates
- shorten many long names: transactionReceipt => receipt, we know it's (transaction)Receipt, no confusion
- removed huge chunks of legacy codes
- some slight behavior change, previously when receipt not found we usually throw an error, now we just return null instead, which makes more sense semantically, and easier to handle.

## Test
All previous tests still pass. 

Although there are a lot of code change, it's scope is limited to "refactor" only. There is no new/modified features, so in theory I shouldn't need to add more tests for it, no matter how many lines are modified.

If this PR broke any feature X, but somehow managed to pass all prev tests, it's not this PR's fault. I will only blame me-in-the-past for not adding adequate tests when adding feature X 💁.

### DIFF
--- a/examples/waffle/e2e/test/dex.test.ts
+++ b/examples/waffle/e2e/test/dex.test.ts
@@ -98,8 +98,6 @@ describe('dex test', () => {
       value: '0x0'
     });
 
-    console.log(receipt)
-
     expect(hexlifyRpcResult(receipt)).deep.eq(hexlifyRpcResult({
       to: TokenA.address,
       from: aliceEvmAddress,
--- a/examples/waffle/e2e/test/dex.test.ts
+++ b/examples/waffle/e2e/test/dex.test.ts
@@ -86,8 +86,7 @@ describe('dex test', () => {
 
     const txHash = swapWithExactSupplyExtrinsic.hash.toHex();
     const tx = await evmProvider.getTransactionByHash(txHash);
-
-    const receipt = await evmProvider.getTxReceiptByHash(txHash);
+    const receipt = await evmProvider.getReceiptByHash(txHash);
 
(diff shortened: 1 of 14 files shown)


examples/waffle/e2e/test/dex.test.ts
```diff
@@ -99,52 +98,51 @@ describe('dex test', () => {
...
       ],
-      blockNumber: hexlifyRpcResult(tx.blockNumber),
-      cumulativeGasUsed: hexlifyRpcResult(receipt.cumulativeGasUsed),
-      effectiveGasPrice: hexlifyRpcResult(receipt.effectiveGasPrice),
+      blockNumber: tx.blockNumber,
+      cumulativeGasUsed: receipt.cumulativeGasUsed,
+      effectiveGasPrice: receipt.effectiveGasPrice,
       status: '0x1',
       type: '0x0',
-      byzantium: true,
```
zjb0807, line 146: https://github.com/AcalaNetwork/bodhi.js/blob/ecdb246879d97dcdb2b8c004712bc519643f219b/eth-rpc-adapter/src/eip1193-bridge.ts#L340

Removing the `byzantium` field causes exception here

shunjizhan: I think it won't cause exception, but let's remove this extra line
Screenshot 2023-02-19 at 13 17 07

zjb0807 (approved):

shunjizhan:
@zjb0807 `confirmation` is also not a property in receipt, so I also removed it

shunjizhan:
ok receipt formatter will auto add confirmation as `null`, so we still need to delete it, but don't need to calculate it

PR #658, opened 2023-02-14
add canonical check when getting receipt from cache

shunjizhan:
## Change
Added canonical check when getting a receipt from cache. 

In #656 we will also add similar check when using `cache.getAllReceiptsAtBlock`, fortunately we only need 1 extra async call to check if that block is canonical, so it shouldn't have too much impact on performance. 

## Test
- prev tests still pass
- no test for fork, no easy way to setup

### DIFF
--- a/eth-providers/src/base-provider.ts
+++ b/eth-providers/src/base-provider.ts
@@ -86,6 +86,7 @@ import {
   checkEvmExecutionError,
   findTxFeeEvent,
   getAllReceiptsAtBlock,
+  FullReceipt,
 } from './utils';
 import { BlockCache, CacheInspect } from './utils/BlockCache';
 import { TransactionReceipt as TransactionReceiptGQL, _Metadata } from './utils/gqlTypes';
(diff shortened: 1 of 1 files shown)


eth-providers/src/base-provider.ts
```diff
@@ -1459,39 +1459,41 @@ export abstract class BaseProvider extends AbstractProvider {
...
+        return logger.throwArgumentError(
+          'blocktag should be number | hex string | \'latest\' | \'earliest\' | \'finalized\' | \'safe\'',
+          'blockTag',
+          blockTag
+        );
       }
     }
   };
 
+  _isBlockCanonical = async (blockHash: string, _blockNumber?: number): Promise<boolean> => {
```
zjb0807, line 1480: Why is `_blockNumber?` here?

shunjizhan:
```
const blockNumber = _blockNumber ?? await this._getBlockNumber(blockHash);
```
if we already know the block number (which is the case for receipt), we can save an async call

zjb0807 (approved):
LGTM

PR #659, opened 2023-02-16
enhanced parseblock and subql

shunjizhan:
#656 is too big and I am gonna break it to a couple smaller self-contained pieces, so it's more reviewable.

## Change
Added a param `targetTxHash` for `getAllReceiptsAtBlock`, so that it can be used to find a single target receipt. Note that the performance should be much better than "first calling getAllReceiptsAtBlock to get all receipts, then filter out the target one".

Although in most cases for RPC we use subql/cache for receipt, a couple methods still require no latency. So we added a new method `getReceiptAtBlockFromChain` in baseProvider to for this. (which is basically what old `getTransactionReceiptAtBlock` does)

Also added a subql method to fetch all receipts at a block, which will be used later.

## Test
still pass

### DIFF
--- a/bodhi/src/Signer.ts
+++ b/bodhi/src/Signer.ts
@@ -269,7 +269,9 @@ export class Signer extends Abstractsigner implements TypedDataSigner {
                   const hex = result.status.isInBlock
                     ? result.status.asInBlock.toHex()
                     : result.status.asFinalized.toHex();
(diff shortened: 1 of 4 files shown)


shunjizhan:
since next PRs depend on this one so I am gonna merge it, but feel free to leave any comments

PR #662, opened 2023-02-20
new getblock flow

### LINKED ISSUE #587: orphan tx support improvement
We seem to have introduced too much complexity by supporting orphan tx, which doesn't happen very often.  Need to do more digging about the performance impact, and some optimization, at least [this](https://github.com/AcalaNetwork/bodhi.js/blob/master/eth-providers/src/base-provider.ts#L1596) since it's not worth looking for orphan tx everytime

### LINKED ISSUE #636: RPC `eth_getBlockByNumber ` failed
```
curl --location --request POST 'https://eth-rpc-acala.aca-api.network/' \
--data-raw '{
    "jsonrpc": "2.0",
    "method": "eth_getBlockByNumber",
    "params": [
        2669090,
        true
    ],
    "id": 1
}'
```

response:
```
{
    "id": 1,
    "jsonrpc": "2.0",
    "error": {
        "code": 6969,
        "message": "Error: Cannot read properties of null (reading 'gas')",
        "_isJSONRPCError": true
    }
}
```

shunjizhan:
### new get block flow
new block flow

### other updates
- fixed an issue that when getting block with `full=false`, can't calculate `gasUsed` for the block
- change "semi" eslint rule to warning, so it won't prevent the code from building in the dev process, we just need to fix the warning before merging

fix #587 
should also fix #636 after subql re-indexing

## Test
all tests still pass, fixed a couple incorrect data in prev tests for `gasUsed`


eth-providers/src/base-provider.ts
```diff
@@ -536,24 +534,25 @@ export abstract class BaseProvider extends AbstractProvider {
...
+    if (receiptsFromSubql?.length) {
+      receipts = receiptsFromSubql.map(subqlReceiptAdapter);
+    } else {
+      /* ----------
+         if nothing is returned from subql, either no tx exists in this block,
+         or the block not finalized. So we still need to ask block cache.
+                                                                    ---------- */
+      receipts = await isCanonical
+        ? this.blockCache.getAllReceiptsAtBlock(blockHash)
+        : [];
```
zjb0807, line 547:
```suggestion
      receipts = this.blockCache.getAllReceiptsAtBlock(blockHash);
```
if `_blockTag` is a forked block hash,  we should return the actual receipts of the block

shunjizhan: at first I implemented that way, https://github.com/AcalaNetwork/bodhi.js/pull/662/commits/ccc11ff6a2bb550cea74879bcda16f92ab8df6a7#diff-8e05d01e1b336d3501a54605915fd1c9116c51bc23417352c54359c4ca197119L544-L548

but after a second thought, that might cause some inconsistency that when use call getblock, they saw transactions in it, but when they call get receipts for these tx, it will return `null`. What do you think?

shunjizhan: This looks insistent to me
```
const block = getBlock(forked-block-hash); 
const firstTx = block.tx[0];
const receipt = getReceipt(firstTx.hash);  // null
```

zjb0807 (approved):

PR #680, opened 2023-03-01
polish

shunjizhan:
## Context
one single async call usually adds 100-500ms in response time, which is already slower than **ALL** sync operations. So in order to achive best performance, we should set the target to "never trigger any redundant async call", and there are many places that we should cache.

#### already implemented
- [x] storage cache
  - [x] `api.query.*`
- [x] receipts cache
- [x] finalized blockhashes
- [x] current finalized block number/hash 

#### cached by subway
- [x] rpc result cache
  - [x] `api.rpc.getBlock`
  - [x] `api.rpc.getHeader`

#### not yet
- [ ] current block number/hash 

Although we do have a cache size limit, in practice most ETH RPC calls target most recent couple blocks (deploy to current block, get current block info, check tx receipt at curernt block ...), which means even only 0.1% of latest blocks are cached, 90% of requests can be served.

As a result, after all these caches are implemented, we should be confident that 90%+ of external ETH RPC calls trigger **ZERO** async calls to external node, and are responded almost instantly! This will also greatly reduce node burden at high traffic time by 100x.

## Change
- keeps track of latest block number and hash
- moved `getTransactionRequest` pure function out of base-provider
- fixed block gaslimit to `29990102`
- fixed a small possible timing issue for `isReady` with local mode. Basically make sure `this.(latest)block(number|hash)` are all ready before `isReady` returns, this might add 100ms for ready time, but that's fine since it only happens once at start.

## Test
still pass


zjb0807 (commented):
LGTM

xlc:
We could have a subway node hosted next to eth-rpc-adapter to cache Substrate RPC responses.
We could also have another subway node hosted next to user to cache Eth RPC responses.

So while it is useful to get the basic cache done, but no need to make it exhaustive.

shunjizhan:
@xlc I removed rpc caches for getBlock and getHeader, and we will handle rpc caches with subway. I think it's still worth tracking latest block number and hash though, it requires no space and is this cacheable (since it's dynamically updating)

xlc (approved):

PR #686, opened 2023-03-10
gas V2 implementation

shunjizhan:
## Change
added implementation for the new gas system. It's shield behind `process.env.V2` for now. 

Whether V2 is on only affects gas **encoding**, i.e. if V2 is on, all new  tx is encoded in new format, else in old format. For **decoding**, it can auto handle both V1/V2 format.

## Test
- all old tests still pass, meaning it's totally compatible with V1 format
- had some manual test with V2 format and https://github.com/AcalaNetwork/Acala/pull/2491, basic contract call/deployment and sending ACA worked.

## Todo
tentative rollout plan (improve bodhi infra and examples by the way):
### 1) **compatible** rollout prepare:
  - build new mandala image after https://github.com/AcalaNetwork/Acala/pull/2491 is merged
  - switch Rush with yarn (maybe)
  - upgrade CI with new image
  - update all examples and tests, remove the gas override parts, aggregate examples and use TS for them
  - update CI to include new tests/examples
###  2) **compatible** rollout
  - runtime upgrade testnets
  - upgrade testnets bodhi with `V2=true`, which uses V2 encoding, and can decode both V1/V2
  - do some testing, ask devs to upgrade testnet Dapps/tools to use new format
  - runtime upgrade mainnets
  - upgrade mainnet bodhi with `V2=true`, which uses V2 encoding, and can decode both V1/V2
  - ask partners to upgrade their dapps/tools to use new format
### 3) **breaking** rollout prepare
  - bodhi and node remove support for old version
  - remove all old tests and examples
###  4) **breaking** rollout
  - runtime upgrade testnets
  - upgrade testnets bodhi, now only supports V2
  - do some testing
  - runtime upgrade mainnets
  - upgrade mainnet bodhi, now only supports V2
  - some final testing


eth-providers/src/base-provider.ts
```diff
@@ -862,21 +879,77 @@ export abstract class BaseProvider extends AbstractProvider {
...
+    return encodeGasLimit(txFee, gasPrice, gasLimit, usedStorage);
+  };
+
+  _estimateGasCost = async (extrinsic: SubmittableExtrinsic<'promise', ISubmittableResult>) => {
+    const header = await this._getBlockHeader('latest');
+    const isLocalModeFirstBlock = this.localMode && header.parentHash.toHex() === ZERO_BLOCK_HASH;
+    const parentHash = isLocalModeFirstBlock
+      ? header.hash
+      : header.parentHash;
+    const apiAtParentBlock = await this.api.at(parentHash);
```
zjb0807, line 941: The current hash should be used here. 
The calculation method of the handling fee has changed, not sure if the calculation is correct.
Keep this in mind during testing.

shunjizhan: I will pay attention to the gas calculation when testing to make sure the it's correct.

Actually as long as it's not too far from the correct gas, it should be fine, since we don't need an exact match here, we usually have `gasLimit = gasEstimation * 120%` to make sure tx pass

zjb0807 (approved):
LGTM

PR #701, opened 2023-04-08
update bodhi multisigner

ermalkaleci:


bodhi/src/MultiSigner.ts
```diff
@@ -0,0 +1,28 @@
+import type { Signer, SignerResult } from '@polkadot/api/types';
+import type { SignerPayloadJSON } from '@polkadot/types/types';
+
+export class MultiSigner implements Signer {
+  #singers: Record<string, Signer>;
```
ermalkaleci, line 5: Typo

shunjizhan (approved):
it would be nice to have some examples/tests for multiSigner later

ermalkaleci:
> it would be nice to have some examples/tests for multiSigner later

All the tests are using utils/getTestUtils so it's basically tested

shunjizhan:
wallets from util is `SubstrateWallet` instead of `multiSigner`, right?

shunjizhan:
but multisigner is not actually used anywhere so should be fine

ermalkaleci:
This will fix the issue with creating multiple BodhiSigner. Instead of replacing signer it will append to multisigner. So now we can create multiple wallets with different signer. Doesn't need to be SubstrateSigner

PR #711, opened 2023-04-19
pack eth-rpc-adapter to reduce docker image

ermalkaleci:
uncompressed image from 3GB to 130Mb


ermalkaleci:
uncompressed image from 3GB to 100Mb

ermalkaleci:
@shunjizhan can you give this a try?

shunjizhan (approved):
tried it out, looks good to me

shunjizhan:
by the way if we are going to use this image in prod, it might be worth to update CI build to use ncc too, which makes prod and test more consistent

PR #714, opened 2023-04-26
added json rpc provider

### LINKED ISSUE #698: export a JsonRpcProvider
this will enable the common usage of 
```ts
const provider = new JsonRpcProvider(ETH_RPC_URL)
```

shunjizhan:
## Context
A standard way to interact with EVM network is through `JsonRpcProvider(ethRpcUrl)`. All network details is encapsulated behind ETH rpc, so that users won't need any "insider's context" for the network. They shouldn't need to care about the underlying architecture: whether it's OP or ZK rollup, substrate, or any other type of EVM network, all they need is the eth rpc url.

However in our current implementation, we need to do `EvmRpcProvider(substrateUrl, subqueryUrl)`, which sometimes (actually often) introduces confusion for devs. They need extra knowledge such as:
- what's the difference between substrate url and eth rpc url?
- what is subquery?
- when do I need to provide subquery url?

## Change
this PR adds a `AcalaJsonRpcProvider` which is compatible with `ethers.JsonRpcProvider`. It solves the known issue that tx hash calculation is different on acala, which makes `ethers.JsonRpcProvider` to throw error.

Now ppl should be able to interact with EVM+ with `AcalaJsonRpcProvider` and eth rpc url, even they have no idea what substrate is, just like how we can interact with optimism without knowing what a sequencer is. 

fix #698 

## Test
Added tests for all basic functionalities of  `JsonRpcProvider`, all of them are compatible

There might be corner cases that aren't tested, we can still roll out this as beta version, and let other devs try it out. If they found any issue we can fix then


shunjizhan:
I think it's ready for review, new tests all passed locally, just need to figure out how to integrate it with CI

PR #723, opened 2023-05-05
make eth call v2 compatible with runtime 2160

shunjizhan:
## Change
Updated deps to use latest types with `polkadot/api@10`, however it still can't find runtime `TransactionPayment v3`, so I had to manually patch it for now.

## Test
manually tested with the following local setup, and the routing process still worked
- v2 rpc adapter connecting to karura testnet, which is runtime 2160
- relayer with AcalaJsonRpcProvider connecting to local eth rpc

### DIFF
--- a/packages/bodhi/package.json
+++ b/packages/bodhi/package.json
@@ -9,10 +9,10 @@
     "build": "tsc --build ./tsconfig.json"
   },
   "peerDependencies": {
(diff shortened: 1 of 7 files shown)


ermalkaleci:
LGTM

ermalkaleci (approved):

PR #726, opened 2023-05-09
use vitest

ermalkaleci:


packages/eth-providers/src/__tests__/e2e/evm-rpc-provider.test.ts
```diff
@@ -25,7 +25,10 @@ describe('initilization', async () => {
   const provider = EvmRpcProvider.from(ACALA_NODE_URL);
   await provider.isReady();
 
-  afterAll(async () => await provider.disconnect());
+  afterAll(async () => {
+    await sleep(5000);
```
shunjizhan, line 29: why do we need to wait here?

packages/eth-providers/src/__tests__/e2e/json-rpc-provider.test.ts
```diff
@@ -21,14 +21,23 @@ describe('JsonRpcProvider', async () => {
...
 
   const provider = new AcalaJsonRpcProvider('https://eth-rpc-karura.aca-api.network');
+  await provider.ready;
   const usdcAddr = '0x1F3a10587A20114EA25Ba1b388EE2dD4A337ce27';
   const usdc = new Contract(usdcAddr, erc20Json.abi, provider);
 
   /* --------- local --------- */
   const testKey = 'a872f6cbd25a0e04a08b1e21098017a9e6194d101d75e13111f71410c59cd57f';   // 0x75E480dB528101a381Ce68544611C169Ad7EB342
   const providerLocal = new AcalaJsonRpcProvider(localEthRpc);
+  await providerLocal.ready;
```
shunjizhan, line 31: for JsonRpcProvider I think we don't need to wait, since all communication is through json rpc

packages/eth-providers/src/__tests__/e2e/json-rpc-provider.test.ts
```diff
@@ -21,14 +21,23 @@ describe('JsonRpcProvider', async () => {
...
   const wallet = new Wallet(testKey, providerLocal);
 
+  afterAll(async () => {
+    await sleep(1000);
+    provider.removeAllListeners();
```
shunjizhan, line 36: for JsonRpcProvider all subscription is handled via polling in interval, there is no websocket subscription, so I think we might also not need these

packages/eth-providers/src/__tests__/e2e/safemode.test.ts
```diff
@@ -112,52 +106,55 @@ describe('safe mode', () => {
...
       subscription: sub,
-      result: {
-        hash: curHash,
-      },
+      result: expect.objectContaining({
+        hash: latestHash,
+      }),
     });
 
-    expect(safeCb).to.have.not.been.calledWithMatch({
```
shunjizhan, line 156: this is to make sure safe rpc cb isn't triggered when it shouldn't

packages/eth-rpc-adapter/package.json
```diff
@@ -7,10 +7,10 @@
   "scripts": {
     "clean": "rm -rf lib tsconfig.tsbuildinfo",
     "build": "tsc --build tsconfig.json",
-    "test:e2e": "SKIP_PUBLIC=true mocha -r tsconfig-paths/register src/__tests__/e2e/*",
+    "test:e2e": "SKIP_PUBLIC=true vitest --run --config vitest.config.e2e.ts",
     "start": "ts-node -r tsconfig-paths/register src/index.ts",
-    "dev": "ts-node-dev -T -r tsconfig-paths/register src/index.ts | pino-pretty --singleLine --colorize --ignore time,hostname,jsonrpc",
-    "test:CI": "mocha -r tsconfig-paths/register src/__tests__/e2e/* --timeout 300000",
+    "dev": "ts-node-dev -T -r tsconfig-paths/register src/index.ts -l | pino-pretty --singleLine --colorize --ignore time,hostname,jsonrpc",
+    "test:CI": "vitest --run --config vitest.config.e2e.ts",
```
shunjizhan, line 13: currently we have two vitest configs, I am wondering if there is any issue if we just use one config for all tests? We usually don't have to need to run { normal, e2e } tests seperately, and in CI we should run all tests

ermalkaleci: regarding e2e I can't find a better away to exclude e2e to sharedConfig and include them on config.e2e

shunjizhan (approved):

PR #728, opened 2023-05-11
refactor provider

ermalkaleci:


packages/eth-providers/src/base-provider.ts
```diff
@@ -1420,70 +1392,32 @@ export abstract class BaseProvider extends AbstractProvider {
 
     result.timestamp = Math.floor((await this.queryStorage('timestamp.now', [], result.blockHash)).toNumber() / 1000);
 
-    result.wait = async (confirms?: number, timeout?: number) => {
+    result.wait = async (confirms?: number, timeout_ms?: number) => {
```
shunjizhan, line 1395: `timeoutMs`?

packages/eth-providers/src/base-provider.ts
```diff
@@ -1420,70 +1392,32 @@ export abstract class BaseProvider extends AbstractProvider {
...
-      return new Promise((resolve, reject) => {
-        const cancelFuncs: Array<() => void> = [];
+      let wait$ = timeout_ms ? this.head$.pipe(
+        timeout({
+          first: timeout_ms,
+          with: () => throwError(() => logger.makeError('timeout exceeded', Logger.errors.TIMEOUT, { timeout: timeout_ms })),
+        })) : this.head$;
 
-        let done = false;
+      wait$ = wait$.pipe(first((head) => head.number.toNumber() - startBlock === confirms));
```
shunjizhan, line 1413:
```suggestion
      wait$ = wait$.pipe(first((head) => head.number.toNumber() - startBlock + 1 === confirms));
```
startBlock is where the tx is mined, so when current block === startBlock, it's 1 confirmation

packages/eth-providers/src/base-provider.ts
```diff
@@ -288,12 +291,25 @@ export abstract class BaseProvider extends AbstractProvider {
...
-  subscriptionStarted: boolean;
+
+  #subscription: Promise<() => void> | undefined;
+  head$: Observable<Header>;
+  finalizedHead$: Observable<Header>;
+
+  readonly queue = new AsyncScheduler(AsyncAction);
+
+  get latestBlockHash(): Promise<string> {
+    return firstValueFrom(this.head$.pipe(map(header => header.hash.toHex())));
```
shunjizhan, line 302: just wanna double check: does `head$` emit current block header immediately on subscription? Otherwise this will return next block hash instead.

Also, I am wondering if this would degrade performance? Previously we cache these 4 variables locally, since they are used very often by all methods, and instead of querying `await api.rpc.chain.getBlockHash()` each time, we hope to achieve an instant query time. 

(correct me if I am wrong) In the Rxjs version, every query to `latestBlockHash` needs to setup a subscription, transform that to promise, resolve it, and close subscription. We basically changed a sync local variable read to an asynchronous fetch, similar to using `await api.rpc.chain.getBlockHash()`, I am not too sure what's the benefits for this?

shunjizhan (approved):
discussed with @ermalkaleci and this shouldn't be impacted by network overhead

packages/eth-providers/vitest.config.e2e.ts
```diff
@@ -7,6 +7,7 @@ export default defineConfig({
     threads: false,
     testTimeout: 300_000,
     hookTimeout: 60_000,
+    dangerouslyIgnoreUnhandledErrors: true,
```
zjb0807, line 10: Is this ignoring something known?

ermalkaleci: Sometimes socket will throw error when we disconnect api. We had this before. I fixed some of them by adding some delay on disconnect but it's still flaky

ermalkaleci: Maybe it's better to keep it flaky. It will remind us to fix it 😅

packages/eth-providers/src/base-provider.ts
```diff
@@ -347,52 +364,101 @@ export abstract class BaseProvider extends AbstractProvider {
...
+      headTasksSub.unsubscribe();
+      finalizedTasksSub.unsubscribe();
     };
   };
 
-  _notifySubscribers = async (blockHash: string, receipts: FullReceipt[]) => {
+  _onNewHead = async ([header, attempts]: [Header, number]) => {
+    attempts--;
+    const blockHash = header.hash.toHex();
+    try {
```
shunjizhan, line 414: curious why we add try catch block here, is there new timing stuff that might make it throw?

packages/eth-providers/src/base-provider.ts
```diff
@@ -546,17 +556,35 @@ export abstract class BaseProvider extends AbstractProvider {
       await this.api.isReadyOrError;
       await this.getNetwork();
 
-      if (!this.subscriptionStarted) {
-        this.subscriptionStarted = true;
-        await this.startSubscriptions();
+      if (!this.#subscription) {
+        this.#subscription = this.startSubscriptions();
```
shunjizhan, line 560: here I am thinking maybe we can unwrap `#subscription` from promise and do this one line
```
this.#subscription ??= await this.startSubscriptions();
```

packages/eth-providers/src/base-provider.ts
ermalkaleci: previously we just called the method and ignored the error. This is implementing a retry with a delay of 1sec

packages/eth-providers/src/__tests__/e2e/json-rpc-provider.test.ts
```diff
@@ -149,10 +148,10 @@ describe('JsonRpcProvider', async () => {
 
       expect(await echo.echo()).to.equal('Deployed successfully!');
 
-      await (await echo.scream('hello Gogeta!')).wait();
```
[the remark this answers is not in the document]
ermalkaleci: ok, will double check. I thought this was an unnecessary wait

packages/eth-providers/src/base-provider.ts
ermalkaleci: the reason I set the promise then wait is because if ready is called twice at the same time it won't call  this.startSubscriptions twice

PR #729, opened 2023-05-12
Fix estimateResources

zjb0807:
Fix https://github.com/AcalaNetwork/Acala/pull/2530


packages/eth-providers/src/base-provider.ts
```diff
@@ -1080,16 +1081,16 @@ export abstract class BaseProvider extends AbstractProvider {
     gasLimit: BigNumber;
     usedStorage: BigNumber;
   }> => {
-    const MAX_GAS_LIMIT = BLOCK_GAS_LIMIT;
+    const MAX_GAS_LIMIT = BLOCK_GAS_LIMIT * 10; // capped at 10x (1000%) the current block gas limit
```
zjb0807, line 1084: https://github.com/AcalaNetwork/Acala/blob/dfbda3a4cc3d6b3ea78bdfdf542b3c2d7fc5d576/modules/evm/rpc/src/lib.rs#L203-L204

packages/eth-providers/src/consts.ts
```diff
@@ -87,7 +87,8 @@ export const ORPHAN_TX_DEFAULT_INFO = {
   ...DUMMY_V_R_S,
 };
 
-export const BLOCK_GAS_LIMIT = 29_990_102;
+export const BLOCK_GAS_LIMIT = 29_990_016;
+export const BLOCK_STORAGE_LIMIT = 3_670_016;
```
zjb0807, line 91: https://github.com/AcalaNetwork/Acala/blob/dfbda3a4cc3d6b3ea78bdfdf542b3c2d7fc5d576/runtime/integration-tests/src/evm.rs#L1158-L1159

zjb0807: We should get this parameter from the chain later

ermalkaleci (approved):

PR #740, opened 2023-05-29
fix metamask send token gaslimit

shunjizhan:
## Context
a long time ago, some random metamask user said: "I tried to transfer token via MM, but failed due to gas underestimation". MM engineers only have 2 hours to solve this issue, so they made a fastest decision: simply 1.5X gasLimit for ALL token transfer, for ALL networks ¯\_(ツ)_/¯

## Change
This occasionally causes our new gas system to fail. For example when bbbcc = 00300, after 1.5X it becomes 00450, which correspond to storageLimit of 2^50.

Since this is a very corner case (only happens when directly transfer token via MM), the easiest workaround is to make sure bbb is even, so after 1.5X it won't affect cc. In the above example, when modify bbbcc from 00300 => 00400, after 1.5X it's 00600, cc is not affected.

## Test
tested locally connecting to karura testnet, and previous failing MM token transfer works

### DIFF
--- a/packages/eth-providers/src/consts.ts
+++ b/packages/eth-providers/src/consts.ts
@@ -89,6 +89,7 @@ export const ORPHAN_TX_DEFAULT_INFO = {
 
 export const BLOCK_GAS_LIMIT = 29_990_016;
 export const BLOCK_STORAGE_LIMIT = 3_670_016;
+export const MAX_GAS_LIMIT_CC = 22;  // log2(BLOCK_STORAGE_LIMIT)
 
 export const ONE_GWEI = 1_000_000_000n;
 export const TEN_GWEI = ONE_GWEI * 10n;
--- a/packages/eth-providers/src/consts.ts
+++ b/packages/eth-providers/src/consts.ts
@@ -89,7 +89,7 @@ export const ORPHAN_TX_DEFAULT_INFO = {
 
 export const BLOCK_GAS_LIMIT = 29_990_016;
 export const BLOCK_STORAGE_LIMIT = 3_670_016;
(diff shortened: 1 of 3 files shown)


packages/eth-providers/src/utils/transactionHelper.ts
```diff
@@ -121,16 +121,28 @@ export const encodeGasLimit = (
...
+    : 0;
+
+  let encodedGasLimit = gasLimit.div(GAS_LIMIT_CHUNK).add(1);
+  if (isTokenTransfer) {
+    // for token transfer, need to make sure bbb is even
+    // so when metamask 1.5X gasLimit, it won't affect cc
+    encodedGasLimit = encodedGasLimit.mod(2).eq(0)
+      ? encodedGasLimit
+      : encodedGasLimit.add(1);
+  }
```
zjb0807, line 139:
```
encodedGasLimit = encodedGasLimit.div(1000).add(1).mul(1000)
```
It can also ensure that the last 3 bits are 0, reducing the impact of multiples

packages/eth-providers/src/consts.ts
```diff
@@ -89,6 +89,7 @@ export const ORPHAN_TX_DEFAULT_INFO = {
 
 export const BLOCK_GAS_LIMIT = 29_990_016;
 export const BLOCK_STORAGE_LIMIT = 3_670_016;
+export const MAX_GAS_LIMIT_CC = 22;  // log2(BLOCK_STORAGE_LIMIT)
```
zjb0807, line 92:
```suggestion
export const MAX_GAS_LIMIT_CC = 21;  // log2(BLOCK_STORAGE_LIMIT)
```

2^22 > BLOCK_STORAGE_LIMIT

zjb0807 (approved):

PR #745, opened 2023-06-07
use truffle tutorial v2

shunjizhan:


shunjizhan:
the only failing test looks like a flaky one

PR #815, opened 2023-09-01
use latest subql cli

shunjizhan:
## Change
- updated to latest subql cli and types
- fixed a flaky test
- updated to latest vitest


shunjizhan:
the only test failure seems to be related to kusuma/karura stopping producing blocks

PR #828, opened 2023-09-20
fix `eth_getLogs` for latest block timing issue

shunjizhan:
## Issue
when query logs for a block as soon as it's finalized, it will return empty logs since subql is still indexing this block. This is usually fine but breaks subquery log handlers, which exactly queries logs for a block as soon as it's finalized.

## Change
adds a check for get logs to make sure it waits until the indexer finished indexing required range, since now we don't support unfinalized logs, so the upper bound is latest finalized log.

## Test
did some manual test locally, and the subql issue seems to be gone

### DIFF
--- a/packages/eth-providers/src/base-provider.ts
+++ b/packages/eth-providers/src/base-provider.ts
@@ -1785,8 +1785,8 @@ export abstract class BaseProvider extends AbstractProvider {
 
     const subqlLogs = await this.subql.getFilteredLogs(filter);   // only filtered by blockNumber and address
     return subqlLogs
-      .filter((log) => filterLogByTopics(log, filter.topics))
-      .map(this.formatter.filterLog);
+      .filter(log => filterLogByTopics(log, filter.topics))
+      .map(log => this.formatter.filterLog(log));
   };
 
   getIndexerMetadata = async (): Promise<_Metadata | undefined> => {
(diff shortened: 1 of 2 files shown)


packages/eth-providers/src/base-provider.ts
```diff
@@ -1774,10 +1775,22 @@ export abstract class BaseProvider extends AbstractProvider {
...
+    // make sure subql already indexed all target blocks, up until the latest finalized block
+    const upperBoundry = await this.finalizedBlockNumber;
+    const targetBlock = filter.toBlock <= upperBoundry
+      ? filter.toBlock
+      : upperBoundry;
+    let lastProcessedHeight = await this.subql.getLastProcessedHeight();
+    while (lastProcessedHeight < targetBlock) {
+      await sleep(1000);
+      lastProcessedHeight = await this.subql.getLastProcessedHeight();
+    }
```
zjb0807, line 1788: If if `toBlock` is very large, will this hang?

shunjizhan: so we have this https://github.com/AcalaNetwork/bodhi.js/pull/828/files#diff-1dc05f5efb9f973bee2a8da85d8839f8f96514217f7336e218b5c2535bd20baeR1780-R1783

shunjizhan: if toblock is larger than latest finalized block, we will use latest finalized block

zjb0807 (approved):

PR #831, opened 2023-09-22
polish `eth_getLogs` to prevent hanging

shunjizhan:
## Change
if subql had any issue and is delayed, `eth_getLogs` could hang for a very long time. So we add a check to throw error directly to prevent hanging.

## Test
added test for this senario, and added some tests for #828

### DIFF
--- a/packages/eth-providers/src/base-provider.ts
+++ b/packages/eth-providers/src/base-provider.ts
@@ -1782,6 +1782,15 @@ export abstract class BaseProvider extends AbstractProvider {
       ? filter.toBlock
       : upperBoundry;
     let lastProcessedHeight = await this.subql.getLastProcessedHeight();
+
+    if (targetBlock - lastProcessedHeight > 3) {
+      return logger.throwError(
+        'subql is not synced with the target block range, please wait for the indexer to catch up.',
+        Logger.errors.SERVER_ERROR,
+        { targetBlock, lastProcessedHeight }
+      );
+    }
+
     while (lastProcessedHeight < targetBlock) {
       await sleep(1000);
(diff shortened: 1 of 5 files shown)


packages/eth-providers/src/base-provider.ts
```diff
@@ -1782,6 +1782,16 @@ export abstract class BaseProvider extends AbstractProvider {
...
+    if (targetBlock - lastProcessedHeight > 3) {
+      return logger.throwError(
+        'subql is not synced with the target block range, please wait for the indexer to catch up.',
+        Logger.errors.SERVER_ERROR,
+        { targetBlock, lastProcessedHeight }
+      );
+    }
+
+    // wait at most 3 * 12 = 36s
     while (lastProcessedHeight < targetBlock) {
```
zjb0807, line 1795: We should actively exit the loop in case there is a problem with subquery

shunjizhan: I think we should just throw error if subquery has issue, since if we exit loop, we will return false result.

shunjizhan: it's normal if subquery is less than 3 blocks delayed (usually it's not delayed at all), but if it's more than 3 blocks delayed, while loop will take very long time, and RPC will timeout anyways, so we just throw explicit error to let user know what's wrong

shunjizhan: if we ever enter this while loop, that means subquery is considered normal (less than 3 blocks delayed)

zjb0807 (approved):

PR #841, opened 2023-09-27
add net_listening

### LINKED ISSUE #253: implement `net_listening`
remix.ethereum.org requires `net_listening` to operate in the Web3 Provider Environment mode

image

```
[1643223015194] DEBUG (3242770 on XPS-15-7590): incoming request
    jsonrpc: "2.0"
    id: 8517
    method: "net_listening"
    params: []
[1643223015196] DEBUG (3242770 on XPS-15-7590): request completed
    id: 8517
    jsonrpc: "2.0"
    error: {
      "code": -32601,
      "data": "The method net_listening does not exist / is not available.",
      "message": "Method not found"
    }
```

shunjizhan:
## Change
added `net_listening` RPC. It seems to be a dummy RPC for us, since I can't think of any senario that we don't want to accept connection.

Should we check RPC health first, and return `false` when unhealthy? But in many cases, most functionalities still work even for unhealthy RPC (such as subquery delayed), I am not sure if we want no connection at all in this case

fix #253 

## Test
added dummy test for it, just to make sure the RPC exist

### DIFF
--- a/packages/eth-rpc-adapter/src/__tests__/e2e/endpoint.test.ts
+++ b/packages/eth-rpc-adapter/src/__tests__/e2e/endpoint.test.ts
@@ -63,6 +63,7 @@ import {
   getCurBlockHash,
   getNonce,
   RPC_URL,
+  net_listening,
 } from './utils';
 
 import {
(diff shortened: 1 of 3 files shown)


zjb0807 (approved):
https://github.com/ethereum/go-ethereum/blob/614804b33c340cd60cbb6087d898b2968b0da320/eth/backend.go#L475

It always returns `true` in ethereum. 

LGTM

PR #876, opened 2023-11-01
avoid unnecessary binary search in gas estimation

shunjizhan:
## Change
Currently the `eth_estimateGas` RPC is slow with a p95 performance of `3.5s`. This is because we need to binary search the best passing `gasLimit`, which calls `eth_call` repeatedly, usually 5-8 times.

However, in most cases if the contract doesn't have special logics to deal with gas, a `gasLimit` that is slightly more than `gasUsed` should make it work. We only need to do binary search if that's not the case.

So we can first check if `usedGas * 1.2` works, to avoid binary search in most cases, which should speed up the gas estimation, and respectively reduce node burden, by 5-8X .

Also removed some gas related legacy codes

Screenshot 2023-11-01 at 17 46 54

## Test
Added tests to make sure it still works for odd gas contract whose passing gaslimit is much higher than actual used gas


xlc (approved):

zjb0807 (approved):

packages/eth-rpc-adapter/src/__tests__/e2e/endpoint.test.ts
```diff
@@ -633,7 +633,7 @@ describe('endpoint', () => {
         blockNumber: '0xa',
         transactionIndex: '0x0',
         // gasPrice: '0x7b501b0da7',
-        gas: '0x1fd91',
+        gas: '0x25a7f',
```
xlc, line 636: should use snapshot testing so it is just one command to update those values

shunjizhan: in this case it's a little tricky since we are testing `to.deep.contain()`, I didn't find something like `toMatchSnaphostshotContain()`

As a work around, I added `bail = 999` so one CI run should print out all outdated value, which should be pretty easy to manually change.

xlc: check the snapshot test helpers I created https://github.com/AcalaNetwork/chopsticks/blob/master/packages/testing/src/check.ts
basically you can preprocess the object (e.g. filter to only keep the fields you want to check) and then assert it

PR #902, opened 2023-12-11
`getlogs` and internal timing polish

### LINKED ISSUE #863: polish `getlogs` when subql syncing last block
instead of waiting for subql to catch, should probably query logs from chain, or extract logs from receipt. This should be much faster

shunjizhan:
## Context
we have a bug that returns false empty for the corner cases when:
1) get log **immediately** on new block, since subql is usually a couple blocks delayed
2) get receipt **immediately** on new block, since cache is still syncing latest block

1\. was temporarily fixed by #828, but it has very poor performance. 
2\. was not fixed yet

To solve the whole issue completely, we need some bottom level refactor, instead of awkward patches like #828 

### core problem
We directly use node best block as eth best block, so we can query for this block **before** subql or cache is ready. In other words, latest block goes faster than latest state.

### solution
We can redefine "latest block" as "latest cached block", so we can always get logs/receipts for the latest block. 

For logs, we can't wait for subql to catch up, which takes 10+ seconds, instead we can make use of cache to query the blocks missed by subql

## Change
- updated blockcache so we can get logs from it, and it keeps track of last cached block info
- updated latest block to return latest cached block
- now we should always be able to get receipts and logs in latest block since:
  - latest block info is guaranteed to be in cache
  - there is a check to make sure no gap between subql and cache

fix #863 

## Test
- updated some tests to wait for eth rpc to report new block before proceeding
- add tests to make sure we can get logs as well as receipts for the latest block as soon as the block is ready


packages/eth-providers/src/base-provider.ts
```diff
@@ -554,11 +573,9 @@ export abstract class BaseProvider extends AbstractProvider {
   isReady = async (): Promise<void> => {
     try {
       await this.api.isReadyOrError;
-      await this.getNetwork();
```
zjb0807, line 557: `getNetwork` is only here, but it has been deleted. Is it no longer needed?

shunjizhan: it's moved to here https://github.com/AcalaNetwork/bodhi.js/pull/902/files#diff-1dc05f5efb9f973bee2a8da85d8839f8f96514217f7336e218b5c2535bd20baeR609

the order was not correct before, since we usually check `get_network` to make sure the rpc is ready. So we should check `api.isReady` in `getNetwork`, not the opposite

zjb0807 (approved):

PR #927, opened 2024-01-16
polished eth call and gas estimation

shunjizhan:
## Change
- update
  - use `safeStorage` for gas estimation, which is slightly bigger than exact used storage. (`(usedStorage + 64) * 1.1`)
  - use latest acala types
- fix
  - fixed a bug that when estimating gas from old blocks, blockhash is not passed to eth call
- cleanup
  - polished eth call error handling by using built in types
  - removed `eth_getEthResources` which is a legacy helper rpc for gas V1

## Tests
updated tests accordingly


packages/eth-providers/src/base-provider.ts
```diff
@@ -1190,10 +1133,11 @@ export abstract class BaseProvider extends AbstractProvider {
       gasLimit = highest;
     }
 
+    const safeStorage = Math.floor((Math.max(usedStorage, 0) + 64) * 1.1);
```
zjb0807, line 1136:
```suggestion
    // Some contracts rely on block height and have different treatments.
    // Transfer use `usedStorage`
    const safeStorage = ethTx.data.length() == 0 ? Math.max(usedStorage, 0) : Math.floor((Math.max(usedStorage, 0) + 64) * 1.1);
```

PR #980, opened 2024-04-05
use latest subql node v3 and subql query v2

### LINKED ISSUE #972: Update `subql/node` and `subql/query` versions
https://github.com/subquery/subql/releases

shunjizhan:
## Change
- updated subql cli and types to latest version
- updated subql node to latest v3, and subql query to latest v2
- use project template to generate project configs, instead of the hard coded ones
- fixed a couple TS errors in mapping handler
- removed subql height strict check. If subql misses some blocks, we will know by health check, and 99% of receipt/log can still be fetched. So we should not throw error for all

fix #972 

## Test
- tested locally and can now index acala block 5899584


zjb0807 (commented):
LGTM

PR #995, opened 2024-04-29
polished batch evm tx parsing

shunjizhan:
## Change
Currently for all extrinsic with multiple evm events, we joins events and make it a single evm tx. 

However, this doesn't make sense for batch evm tx, where each of the evm event is a real evm tx. For example our lottery use batch "approve + draw", and the `draw` transactions cannot be found at all.

So for batch evm tx, we should assign each of the event a unique tx receipt. This can be done simply by treating them as orphan events, since they don't own a 1-to-1 tx hash.

## Test
- added tests for the lottery batch `approve + draw` tx to make sure two separate receipts can be parsed correctly.
- other transactions can still be parsed, since all old tests still pass


packages/eth-providers/src/utils/utils.ts
```diff
@@ -330,6 +331,11 @@ export const checkEvmExecutionError = (data: CallReturnInfo): void => {
 
 export const isEvmExtrinsic = (e: Extrinsic): boolean => e.method.section.toLowerCase() === 'evm';
 
+export const isBatchExtrinsic = (extrinsic: GenericExtrinsic<AnyTuple>): boolean => {
+  const { method: { method, section } } = extrinsic;
+  return section.toString() === 'utility' && method.toString().includes('batch');
```
xlc, line 336: this doesn't cover `batchAll` and `forceBatch`
another way is to check if there is `utility.batchCompleted` or `utility.batchCompletedWithErrors` event

shunjizhan: how about `method.toString().toLowerCase().includes('batch')`?

xlc: that could work. but it is still better to use events instead of checking the extrinsic name. otherwise for example it won't be able to deal with proxy + batch call

xlc (approved):

zjb0807 (approved):

PR #1015, opened 2024-09-09
update deps

ntduan:


shunjizhan:
somehow new pjs introduces a bug #1020 , it's a relative corner case so we can merge this for now so new runtime testing to proceed

PR #1028, opened 2024-09-26
fix and enable eth provider tests

shunjizhan:


shunjizhan:
@xlc it's odd that chopsticks@0.6.0 [broke the CI](https://github.com/AcalaNetwork/bodhi.js/actions/runs/11066466113/job/30810063605), and I had to [revert to 0.5.0](https://github.com/AcalaNetwork/bodhi.js/pull/1028/commits/9a72f76279e3e87a15ff1a472a980ffec0c6473d) for it to work.

It might not be a bug but some weird compatibility issue among { bun, chopsticks, github-ci }, so just take a note here in case something similar happen

xlc:
maybe it is related to https://github.com/AcalaNetwork/chopsticks/pull/826

Issue #332, opened 2022-03-31
Improved concurrency performance


ntduan:
https://github.com/polkadot-js/api/issues/4707

shunjizhan:
I think we can also do some caching for `apiAt` for the latest 10 blocks, since most queries ask for these blocks. The implementation should be pretty similar to how we cache tx hashes

shunjizhan:
so basically if many request comes in, they can share the same instance of `apiAt` and won't need to make many new connections. But still, this is a workaround we can do, it will definitely be nice to improve concurrency performance from the api itself

ntduan:
> I think we can also do some caching for `apiAt` for the latest 10 blocks in RPC adaptor, since most queries ask for these blocks. The implementation should be pretty similar to how we cache tx hashes

Yes, the api has fixed some of the problems. But it still creates a new api object for each blockhash, which can cause a lot of memory consumption and service crash when processing many requests at the same time. I will create a block registry based cache.

PR #1039, opened 2024-10-31
cache polish

shunjizhan:
## Change
There seems to be a bug with our storage query helper, which is very random and cannot be reproduced (so far). Before we put too much bandwidth catching the tricky bug, we can use a slightly less optimized but more stable approach: classy way to query storage with apiAt.

Also:
- fixed a bug in block author calculation.
- added a couple new caches for common calculations

## Test
All tests still passed, also added tests to make sure cached request are much faster

## Performance Analysis
These analysis are mostly in theory, we can see how it actually goes by taking a snapshot of current RPC performance, and compare.

There are two main steps for a storage call:
1) construct apiAt
2) make the storage call

### step 1
apiAt construction [already has internal cache](https://github.com/polkadot-js/api/blob/9becdced121977a22c81df4d037c9d0e60b370a5/packages/api/src/base/Init.ts#L168-L186) for `runtimeVersion => registries`, so the only repeated part is the decoration process. These local calculations should take much less time than waiting for the async call in step 2 (TBD).

We can still alleviate repeated decoration by caching `blockHash => apiAt`. In practice most of the calls should be targeting recent blocks, caching only 100 apiAt could already have a sufficiently great hit rate, while maintaining a low memory footprint.

### step 2
We don't necessarily need to cache storage data, since the hit rate will likely be low. Each storage call queries for a specific block with very specific params, such as "nonce for address XXX at block YYY", which is not very likely to repeat. 

On the other hand, cacheing `blockHash => *` should have a much better hit rate, such as `blockHash => apiAt` mentioned above. We should also cache the (very) few common queries that are heavily repeated:
- `blockNumber => blockHash`
- `blockHash => blockData`
- `blockHash => header` (implicitly cached `blockHash => blockNumber`)


xlc (approved):

packages/eth-providers/src/base-provider.ts
```diff
@@ -523,67 +517,6 @@ export abstract class BaseProvider extends AbstractProvider {
...
-    }
-
-    const result = registry.registry.createTypeUnsafe(outputType, [input], {
-      blockHash,
-      isPedantic: !entry.meta.modifier.isOptional,
-    });
-
-    return result as any as T;
-  };
-
```
zjb0807, line 586: https://github.com/AcalaNetwork/bodhi.js/issues/332#issuecomment-1085196293

Not sure if this will cause a timeout in getting Metadata.  Blockscout will query the account balances of many block hashes. We can run Blockscout for a few days to test it.

@ntduan please review this

shunjizhan: fetching metadata in parallel should already be improved https://github.com/polkadot-js/api/pull/4708, but yeah we can try and see how it actually goes

shunjizhan:
an ideal way would still be querying storage directly, but better through api itself https://github.com/polkadot-js/api/issues/6015

PR #1051, opened 2024-11-11
use query storage helper

### LINKED ISSUE #1059: JavaScript heap out of memory on v2.9.4
run with `RPC_CACHE_CAPACITY=100`

```
<--- Last few GCs --->

[1:0x7fc15179c020] 462423328 ms: Scavenge 1946.4 (2080.5) -> 1940.8 (2080.3) MB, 10.0 / 0.0 ms  (average mu = 0.921, current mu = 0.917) allocation failure;
[1:0x7fc15179c020] 462424307 ms: Scavenge 1947.9 (2080.3) -> 1942.2 (2080.3) MB, 13.0 / 0.0 ms  (average mu = 0.921, current mu = 0.917) allocation failure;
[1:0x7fc15179c020] 462426302 ms: Scavenge 1949.6 (2080.9) -> 1943.3 (2096.5) MB, 15.0 / 0.0 ms  (average mu = 0.921, current mu = 0.917) task;


<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

shunjizhan:
## Change
Using api to query storage still has the memory issue https://github.com/polkadot-js/api/issues/5981, so we switch back to the original query storage helper, but removed all cache for storage query. This is because there was a bug in the cache, and as we mentioned in #1039, caching storage won't be too useful.

We still keep the `blockHash => { blockNumber, Header, BlockData }` cache introduced in #1039 , which are commonly used. We keep the apiAt cache as well, which can speed up runtime calls.

fix #1059 

Although there is a small inconsistency when querying `system.account` compared to `apiAt.query`, it does not affect our use case (we have been using this for a long time anyways). But in the future it would be nice to make it better, maybe @ntduan can take a look at #1060 when you got a chance?

## Test
all tests still pass (except the skipped one for storage query helper)

### DIFF
--- a/packages/eth-providers/src/__tests__/BlockCache.test.ts
+++ b/packages/eth-providers/src/__tests__/BlockCache.test.ts
@@ -2,7 +2,7 @@ import { TransactionReceipt } from '@ethersproject/abstract-provider';
 
 import { BlockCache } from '../utils/BlockCache';
 import { describe, expect, it } from 'vitest';
-import { mockChain } from './utils/testUtils';
+import { mockChain } from './utils';
 
 const sortReceipt = (r1: TransactionReceipt, r2: TransactionReceipt) => {
   if (r1.blockNumber !== r2.blockNumber) {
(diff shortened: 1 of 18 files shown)


zjb0807 (approved):
I don't seem to see the changes related to #1052

shunjizhan:
yeah #1052 is not related, it might be some docker issue, let's keep that issue open for now

PR #1063, opened 2025-02-20
fix eth_call with gasLimit

zjb0807:
```
curl --location 'https://eth-rpc-tc9.aca-staging.network' \
--data '{
    "jsonrpc": "2.0",
    "method": "eth_call",
    "params": [
        {
            "data": "0x95d89b41",
            "gas": "0x2faf080",
            "to": "0xbcede5fd04cd6c08ca19d2f5193d6211ab3af5b7"
        },
        {
            "blockHash": "0xc422e7456709a217d364c2db0a3f202a9b6df02794ab5ca141e113719909b6f5"
        }
    ],
    "id": 1
}'
```
`{"jsonrpc":"2.0","error":{"code":-32603,"message":"execution error: outOfGas"},"id":1}`

If the eth_call request has a `gasLimit`, `substrateGas.gasLimit` is 0

### DIFF
--- a/packages/eth-providers/src/base-provider.ts
+++ b/packages/eth-providers/src/base-provider.ts
@@ -772,6 +772,7 @@ export abstract class BaseProvider extends AbstractProvider {
     const callRequest: SubstrateEvmCallRequest = {
       ...ethReq,
       value: ethReq.value?.toBigInt(),
+      gasPrice: ethReq.gasPrice?.toBigInt(),
       gasLimit: ethReq.gasLimit?.toBigInt(),
     };
 
--- a/packages/eth-providers/src/base-provider.ts
+++ b/packages/eth-providers/src/base-provider.ts
@@ -766,14 +766,26 @@ export abstract class BaseProvider extends AbstractProvider {
       this._getBlockHash(blockTag),
     ]);
 
-    ethReq.gasPrice ??= await this.getGasPrice();
-    ethReq.gasLimit ??= BigNumber.from(999999920);
+    let gasLimit;
+    let storageLimit;
+
+    if (!ethReq.gasLimit) {
+      ethReq.gasPrice ??= await this.getGasPrice();
+      ethReq.gasLimit ??= BigNumber.from(999999920);
+
+      const substrateGas = this._getSubstrateGasParams(ethReq);
+      gasLimit = substrateGas.gasLimit;
+      storageLimit = substrateGas.storageLimit;
+    } else {
+      gasLimit = ethReq.gasLimit?.toBigInt();
+      storageLimit = BLOCK_STORAGE_LIMIT;
+    }
 
     const callRequest: SubstrateEvmCallRequest = {
       ...ethReq,
       value: ethReq.value?.toBigInt(),
-      gasPrice: ethReq.gasPrice?.toBigInt(),
-      gasLimit: ethReq.gasLimit?.toBigInt(),
+      gasLimit,
+      storageLimit
     };
 
     const res = await this._ethCall(callRequest, blockHash);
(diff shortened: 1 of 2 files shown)


shunjizhan:
We should pass encoded `gasLimit` to `eth_estimateGas`, otherwise it will be inconsistent with other scenarios.

There are already some tests for overrides: https://github.com/AcalaNetwork/bodhi.js/blob/900dd40fe1f87367ccbc99712daa180c1e6dbcb3/packages/eth-rpc-adapter/src/__tests__/endpoint-tests/eth_estimateGas.test.ts#L46

shunjizhan:
when we call `eth_estimateGas`, it's basically asking "will this request work if I use this gasLimit", and if it does not throw error, we should be able to use the same `gasLimit` for `eth_call` and still work, so in both case it should be encoded

shunjizhan:
If we use eth style gasLimit for `eth_call`, but encoded gaslimit for transaction, this will be very messy and inconsistent.

For example, if we use `"gas": "0x2faf080"` and the request passed `eth_call`, but the same gasLimit won't work when we actually send transaction (since we require encoded gaslimit).

The very point of `eth_call` is to check "if I send this request, will this work?". And in our system, eth style gaslimit will not work, so `eth_call` is expected to fail

shunjizhan:
yeah if we just ignore gasLimit, it might work. But there is still one issue, if we `eth_call` an **invalid** gasLimit (such as low storageLimit), it's expected to fail, so we know this tx won't work. But if we just ignore it, user will say, hey this tx should succeed and send it, then it will actually fail...

shunjizhan:
essentially we should keep the invariant:
- if the tx will work, it should pass `eth_call` and `eth_estimateGas`
- if the tx will fail, it should fail `eth_call` and `eth_estimateGas`
- vise versa

zjb0807:
In my mind, `eth_call` is only used to query contract data and does not determine whether the transaction is successful. 

`eth_estimateGas` will return whether the transaction is successful, and return the `gasLimit`

The order in which the user calls the RPCs should be 
1. maybe call `eth_call` to query contract data
2. call `eth_estimateGas` to get the gasLimit
3. call `eth_sendRawTransaction` to send tx.

Modifying the `eth_call` should not affect `eth_estimateGas` RPC

shunjizhan:
for example for this token transfer tx, if we call the original, it will return successful. But if we change the params such adding `gas: 1`, it will assume we send the tx with `gas: 1`, then return error
```
curl --location 'https://eth.llamarpc.com' \
--data '{
  "id": 0,
  "jsonrpc": "2.0",
  "method": "eth_call",
  "params": [        {
            "data": "0xa9059cbb000000000000000000000000016606acc6b0cfe537acc221e3bf1bb44b4049ee0000000000000000000000000000000000000000000000000000000006422c40",
            "to": "0xdAC17F958D2ee523a2206206994597C13D831ec7",
            "from": "0xA477B7B855366C7738d6d5B86620a0E031d8a3A7"
        },
        "0x14DF8C2"
  ]
}'
```

shunjizhan:
actually, even for view methods, it still consume gas! 

This will also fail (fetching erc20 symbol)

```
curl --location 'https://eth.llamarpc.com' \
--data '{
  "id": 0,
  "jsonrpc": "2.0",
  "method": "eth_call",
  "params": [        {
            "data": "0x95d89b41",
            "to": "0xdAC17F958D2ee523a2206206994597C13D831ec7",
            "from": "0xA477B7B855366C7738d6d5B86620a0E031d8a3A7",
            "gas": 1
        },
        "0x14DF8C2"
  ]
}'
```

zjb0807:
use `GRAPH_ETH_CALL_NO_GAS` to resolve the case.

Issue #1052, opened 2024-11-12
eth-rpc-adapter server should restart automatically 


zjb0807:
When `eth-rpc-adapter` exits or runtime is upgraded, the docker will not restart automatically.  But it was working before

zjb0807:
```
{"level":20,"time":1741903588772,"pid":1,"hostname":"e553cb1c05c7","dd":{"trace_id":"5149450522391938828","span_id":"8013625915150131687","service":"eth rpc","version":"2.9.5"},"0":{"id":0,"jsonrpc":"2.0","method":
"eth_getBlockByNumber","params":["latest",false]},"msg":"⬇ request"}
2025-03-13 22:10:00          API-WS: disconnected from wss://karura-internal-rpc.aca-api.network:8443: 1006:: Abnormal Closure
2025-03-13 22:10:00        RPC-CORE: getHeader(hash?: BlockHash): Header:: disconnected from wss://karura-internal-rpc.aca-api.network:8443: 1006:: Abnormal Closure
2025-03-13 22:10:00        RPC-CORE: getBlock(hash?: BlockHash): SignedBlock:: disconnected from wss://karura-internal-rpc.aca-api.network:8443: 1006:: Abnormal Closure
2025-03-13 22:10:01        RPC-CORE: getBlock(hash?: BlockHash): SignedBlock:: disconnected from wss://karura-internal-rpc.aca-api.network:8443: 1006:: Abnormal Closure
2025-03-13 22:10:01        RPC-CORE: getHeader(hash?: BlockHash): Header:: disconnected from wss://karura-internal-rpc.aca-api.network:8443: 1006:: Abnormal Closure                                                  2025-03-13 22:10:02        RPC-CORE: getBlock(hash?: BlockHash): SignedBlock:: disconnected from wss://karura-internal-rpc.aca-api.network:8443: 1006:: Abnormal Closure
2025-03-13 22:10:02        RPC-CORE: getHeader(hash?: BlockHash): Header:: disconnected from wss://karura-internal-rpc.aca-api.network:8443: 1006:: Abnormal Closure                                                  2025-03-13 22:10:03        RPC-CORE: getBlock(hash?: BlockHash): SignedBlock:: disconnected from wss://karura-internal-rpc.aca-api.network:8443: 1006:: Abnormal Closure
[...]

PR #1064, opened 2025-03-14
Exiting process when ws disconnected

zjb0807:
Closes: #1052 

It seems that dd-trace affects the exit status code


zjb0807:
@shunjizhan Do you have any idea?

shunjizhan:
I am not sure what's the cause, you mentioned it worked before? If so, we might be able to do a binary search with the versions to check from which commit it stopped working

